PR #2, opened 2020-03-15
Attempt to mitigate timeout problem with automat.

cbizon:
Following discussion here:
https://github.com/encode/httpx/issues/411


patrickkwang:
I think this is a bad idea generally, because KPs could be doing who-knows-what and now that we're only allowing 5 jobs at a time, this will just stall a job. If in practice it works well with Automat, we can remove the timeout for the prototype demo.

cbizon:
I think it's a terrible idea.  I'm only proposing it to allow strider to operate during the demo phase.  In 1b we remove it and focus on resolve the underlying  problem with automat.  

However, even if taking the timeout away is not good, we can at least increase it?

PR #17, opened 2020-12-01
Fixed Docker and set up CI.

alongreyber:
Updated Dockerfile Python version.
Created a dev docker-compose file to detect code changes.
Added to requirements.txt.
Created a CI workflow to publish a docker image on releases.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
-FROM python:3.8.1-buster
+FROM python:3.9
+
+# Add image info
+LABEL org.opencontainers.image.source https://github.com/ranking-agent/strider
 
 # install basic tools
 RUN apt-get update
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -13,23 +16,16 @@ RUN useradd -m -u $UID -g $GID -s /bin/bash murphy
...
-ADD --chown=murphy:murphy ./strider /home/murphy/strider
-ADD --chown=murphy:murphy ./setup.py /home/murphy/setup.py
-RUN pip install -e .
-
-# get meta-stuff
-ADD --chown=murphy:murphy ./logging_setup.yml /home/murphy/logging_setup.yml
-ADD --chown=murphy:murphy ./run_workers.py /home/murphy/run_workers.py
-ADD --chown=murphy:murphy ./main.sh /home/murphy/main.sh
+# Copy in files
+ADD --chown=murphy:murphy . .
```
patrickkwang, line 23: There is a bunch of stuff here that the container doesn't need, like /examples and /tests. None of that is huge, but could we exclude them simply by adding to .dockerignore?

patrickkwang (approved):

PR #23, opened 2020-12-08
Redis result storage + logging

alongreyber:


tests/test_strider.py
```diff
@@ -6,11 +6,16 @@
 
 import pytest
 
-from strider.fetcher import StriderWorker
+# Set redis host before importing
+os.environ["REDIS_HOST"] = "fakeredis"
```
patrickkwang, line 10: Is there a more elegant way to handle this switch, or test configuration in general? Is this sort of thing what `conftest.py` is for?

strider/storage.py
```diff
@@ -0,0 +1,99 @@
...
+host = os.getenv('REDIS_HOST', 'redis')
+
+if host == 'fakeredis':
+    import fakeredis
+    r = fakeredis.FakeRedis(
+        encoding="utf-8",
+        decode_responses=True,
+    )
+else:
+    import redis
```
patrickkwang, line 16: I'm a fan of aioredis, but we couldn't async-ify logging without ditching the standard library, which I don't want to do.

strider/worker.py
```diff
@@ -31,14 +31,14 @@ class Worker(ABC):
...
-            queue: asyncio.PriorityQueue,
             num_workers: int = 1,
             **kwargs,
     ):
         """Initialize."""
-        self.counter = kwargs.get('counter', itertools.count())
         self.num_workers = num_workers
-        self.queue = queue
+
+        self.queue = asyncio.PriorityQueue()
```
patrickkwang, line 40: Good call internalizing this 😁.

strider/server.py
```diff
@@ -46,78 +49,110 @@
...
 async def async_query(
-        query: Query,
-        support: bool = True,
-) -> str:
-    """Handle asynchronous query."""
-    query_id = await execute_query(
-        query.message.query_graph.dict(),
-        support=support,
-        wait=False,
+        background_tasks: BackgroundTasks,
```
patrickkwang, line 113: This is cool! I hadn't used BackgroundTasks before.

strider/server.py
```diff
@@ -46,78 +49,110 @@
...
+    kgraph = RedisGraph(f"{qid}:kgraph")
+    results = RedisList(f"{qid}:results")
+    logs = RedisList(f"{qid}:log")
+
+    qgraph.expire(store_results_for)
+    kgraph.expire(store_results_for)
+    results.expire(store_results_for)
+    logs.expire(store_results_for)
+
+    return ReasonerResponse(
```
patrickkwang, line 89: I've tried to avoid instantiating these pydantic models so that the internal logic is not beholden to them. FastAPI will convert dict outputs for us, so I'd just make the dict here. Of course, my strategy does not play nicely with the type hints using pydantic and pyright... so I'm not sure what the right solution is.

strider/server.py
```diff
@@ -46,78 +49,110 @@
...
+async def sync_query(
+        query: Query = Body(..., example=EXAMPLE)
+) -> ReasonerResponse:
+    """Handle synchronous query."""
+    # Generate Query ID
+    qid = str(uuid.uuid4())[:8]
+
+    # Save query graph to redis
+    qgraph = RedisGraph(f"{qid}:qgraph")
+    qgraph.set(
```
patrickkwang, line 147: I just realized that we built an ORM, and that makes me feel icky. We access the query graph a lot; how much overhead is involved in querying redis every time? Since we basically never edit/delete things, can we cache the `get`s using e.g. lru?

cbizon: The reason there are so many ORMs is that they are constantly being spun off of other projects...

tests/test_strider.py
alongreyber: FastAPI has a settings object: [https://fastapi.tiangolo.com/advanced/settings/](https://fastapi.tiangolo.com/advanced/settings/)

We can define default values in a `config.py` file. FastAPI reads environment variables at startup and will override the settings using those. In addition, settings are mutable so we can do `settings.redis_host = 'fakeredis'` in the test file. That should be cleaner than overriding environment variables.

strider/server.py
alongreyber: Yeah that's tough because we will never be able to make pyright happy without using the models, but also there is a lot of work to get all of the internal data structures to use pydantic models. I think for now we should be consistent and use dicts everywhere, then maybe at some point move to the pydantic models.

strider/server.py
alongreyber: I don't think we should worry about the overhead until we have tested and confirmed that it is an issue. Redis is in-memory and designed for very high throughput. We also have avenues for improvement such as caching using LRU. I don't think we need to implement those yet. 

As an aside, I think that we ended up with an ORM because this project benefits from having one. It's intuitive to structure web projects as an application layer interacting with a database.

tests/test_strider.py
patrickkwang: This sounds like a good thing. Does it solve the problem here where the configuration must be interleaved with the imports?

strider/server.py
patrickkwang: Sounds like a plan.

strider/server.py
patrickkwang: Okay.

tests/test_strider.py
alongreyber: Unfortunately it does not. We're creating the redis database connection at import time, so we still have to set the settings and then run the import. I looked at alternatives to this. The FastAPI recommended way is to use their dependency injection/management infrastructure. Reading here if you're interested:

- [Dependency Intro](https://fastapi.tiangolo.com/tutorial/dependencies/)
- [Dependencies in testing](https://fastapi.tiangolo.com/advanced/testing-dependencies/)
- [Database testing](https://fastapi.tiangolo.com/advanced/testing-database/)

I'm not a fan of this structure at all. The TL;DR is we would have to add a dependency parameter to every route, and then pass the database object to every function that needs it. Also, the dependency parameter then shows up in the generated documentation which seems like a very confusing thing to have in our API docs.

patrickkwang: Yeah, that doesn't sound fun, though I thought they were excluded from the API docs. Another option is to wrap the redis things in a generator of some sort so that the module can be imported before the classes are built. But that sounds kludgy.

patrickkwang (approved):

PR #33, opened 2021-01-12
Portal tests

alongreyber:
Added tests for the KnowledgePortal class. This includes tests for the map_prefixes function as well as the wrapped fetch method.


tests/helpers/context.py
```diff
@@ -66,6 +66,26 @@ async def norm_overlay(url):
         yield
 
 
+@asynccontextmanager
+async def response_overlay(url, response: Response):
```
patrickkwang, line 70: This is nice :D

tests/compatibility/test_compatibility.py
```diff
@@ -0,0 +1,227 @@
...
+
+@pytest.mark.asyncio
+@with_response_overlay(
+    settings.normalizer_url,
+    JSONResponse(
+        content={"detail": "No matches found for the specified curie(s)"},
+        status_code=404
+    )
+)
+async def test_normalizer_not_available():
```
patrickkwang, line 129: This gets at some more complex issues. It _might_ be nice to carry on when the normalizer is down, but it's also going to mean we lose a lot of results. We should at least raise a warning somewhere that a critical bit of infrastructure is not working as expected. I'm not sure what mechanism we use exactly.

patrickkwang (approved):

tests/compatibility/test_compatibility.py
alongreyber: That is a good point. I'll add a warning message to the returned log when we can't contact the normalizer.

patrickkwang (approved):

PR #39, opened 2021-01-18
Logging parameter

alongreyber:
Added a parameter for log level for the /query and /query_result endpoints. The default level is error. Also created better error handling for KP requests.


patrickkwang (changes requested):
A bunch of things happened here in addition to adding the logging parameter. Can you please split this into multiple PRs addressing the individual changes?

alongreyber:
I wasn't sure exactly what you wanted to see split into a separate PR. The only thing I found that was easily removable was the change to the KP error handling.

PR #41, opened 2021-01-19
Better KP Error Handling

alongreyber:
Switched to catching httpx.HTTPResponseError and httpx.RequestError for handling KP errors. This allows us to continue running even if there is an error while contacting a KP. The response body is included in the log message.


patrickkwang (changes requested):
This got weird in a way that switching the base branch did not fix, I think because the PR includes a merge commit. Can you rebase the new commits onto v1.0 instead, and force push?

strider/util.py
```diff
@@ -85,23 +81,8 @@ async def post_json(url, request):
...
-            )
-        LOGGER.debug(
-            "%s\n%s\n\x1b[32m%d Success:\x1b[0m %s",
-            url,
-            str(request),
-            response.status_code,
-            response.text,
-        )
-    return response.json()
+        response.raise_for_status()
```
patrickkwang, line 84: Does this obscure the error details that may come from the KP?

alongreyber: No it does not. The details are still accessible using error.response.text.

patrickkwang: Can we add a test to ensure that the body of e.g. a 500 response from a KP is included in our log?

alongreyber: That is already included in the test: https://github.com/ranking-agent/strider/pull/41/files#diff-c80b6c5e90d69579be550bf44a0a9d5f2ea9d4337d3397fa54e535a72107cdf2R187

patrickkwang: It seems to me this confirms that it is a 500 response, but not that any further details are captured.

alongreyber: This checks that we have captured the text in the response from the server, in this case "Internal server error". This can be verified by changing the text in the decorator for the response overlay and will cause the test to fail. Are there any other details you had in mind to capture?

patrickkwang: Ah, got it. Thanks for walking me through that.

patrickkwang (approved):

PR #42, opened 2021-01-19
Test Refactoring

alongreyber:
Moved the load_kps method into the utilities file.

### DIFF
--- a/tests/helpers/utils.py
+++ b/tests/helpers/utils.py
@@ -1,7 +1,23 @@
 import itertools
 import time
+import json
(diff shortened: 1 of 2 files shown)


patrickkwang:
There is an annoying GitHub [bug(?)](https://github.com/isaacs/github/issues/750) where the diff here does not update when the base branch is changed. I'm going to mess around and see if I can get it to recompute.

patrickkwang:
Switching the base branch and back works 👍

patrickkwang (approved):

PR #45, opened 2021-01-20
Pinned Node Fixes

alongreyber:
- Added tests to make sure that we handle pinned nodes properly
- Changed generate_plan to generate_plans
- Modified generate_plans to properly handle nodes that aren't connected to any other nodes
- Added a utility to help make tests with query graphs more concise using the Mermaid syntax

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -267,12 +267,18 @@ async def find_valid_permutations(
     # Filter down node categories using those
     # we have recieved from the KP registry
     # to limit the number of permutations.
-    #
-    # Also filter edge predicates as well.
-    for node in expanded_qg['nodes'].values():
+
+    # We only filter nodes that are touching at least one edge ("connected")
+    connected_nodes = []
+    for edge in expanded_qg['edges'].values():
+        connected_nodes.append(edge['subject'])
+        connected_nodes.append(edge['object'])
+    for node in connected_nodes:
         if 'category' not in node:
             continue
         node['category'] = list(node.pop('filtered_categories'))
+
+    # Also filter edge predicates as well
     for edge in expanded_qg['edges'].values():
         edge['predicate'] = list(edge.pop('filtered_predicates'))
 
@@ -366,8 +372,10 @@ async def generate_plans(
         # Starting at each pinned node, construct a plan
         for pinned in pinned_nodes:
             path_nodes, path_edges = dfs(current_qg, pinned)
-            # If we don't traverse every node we can't use this
-            if set(path_nodes) != set(current_qg['nodes'].keys()):
+            # If we don't traverse every unbound node we can't use this
+            unbound_nodes = [
+                nid for nid, node in current_qg['nodes'].items() if 'id' not in node]
+            if set(unbound_nodes) > set(path_nodes):
                 continue
             # If we don't traverse every edge we can't use this
             if set(path_edges) != set(current_qg['edges'].keys()):
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -357,7 +357,7 @@ async def generate_plans(
     ]
 
     if len(filtered_qg_list) == 0:
-        logger.error({
+        logger.warning({
             "code": "QueryNotTraversable",
             "message":
                 """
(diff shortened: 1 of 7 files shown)


strider/query_planner.py
```diff
@@ -358,23 +364,29 @@ async def generate_plan(
...
+    plans = []
 
     for current_qg in filtered_qg_list:
         # Starting at each pinned node, construct a plan
         for pinned in pinned_nodes:
             path_nodes, path_edges = dfs(current_qg, pinned)
-            # If we don't traverse every node we can't use this
-            if set(path_nodes) != set(current_qg['nodes'].keys()):
+            all_nodes = current_qg['nodes'].keys()
+            bound_nodes = [
```
patrickkwang, line 376: Is `bound_nodes == pinned_nodes`?

strider/query_planner.py
```diff
@@ -388,21 +400,19 @@ async def generate_plan(
                         **op._asdict(),
                         **kp,
                     })
+            plans.append(plan)
 
-    # No steps in the plan here means we didn't find any permutations where
-    # the plan covers all nodes and edges
-    if len(plan) == 0:
+    if len(plans) == 0:
```
patrickkwang, line 405: We're handling this case in two places, here and [there](https://github.com/ranking-agent/strider/pull/45/files#diff-618b9679423647529785427fae14a18324c4029b0e01ba24802812b3c0b5ebd1R134). We should pick one.

tests/query_planner/test_query_planner.py
```diff
@@ -177,6 +178,76 @@ async def test_plan_ex1():
...
+    This should be valid because we only care about
+    a path from a pinned node to all unbound nodes.
+    """
+
+    qg = query_graph_from_string(
+        """
+        n0(( id MONDO:0005148 ))
+        n1(( category biolink:Drug ))
+        n0-- biolink:treated_by -->n1
+	n2(( id MONDO:0011122 ))
```
patrickkwang, line 196: Is this misaligned?

tests/query_planner/test_query_planner.py
```diff
@@ -177,6 +178,76 @@ async def test_plan_ex1():
...
+    Test Pinned -> Unbound + Pinned -> Unbound
+    This should be invalid because there is no path from
+    a pinned node to all unbound nodes.
+    """
+    qg = query_graph_from_string(
+        """
+        n0(( id MONDO:0005737 ))
+        n1(( category biolink:Drug ))
+        n0-- biolink:treated_by -->n1
+	n2(( id MONDO:0011122 ))
```
patrickkwang, line 241: Misaligned?

patrickkwang (changes requested):

strider/query_planner.py
alongreyber: Yes it is. I'll switch to pinned_nodes for consistency in language.

strider/query_planner.py
alongreyber: This is weird, but I'll try to explain my reasoning. Finding zero plans isn't really something we should throw an exception for from generate_plans because no plans is still a valid result. However, we still want to know what went wrong and there's no way to tell exactly what went wrong when we handle it in `fetcher.py`. That is why we're logging the error in the generate_plans function and then handling it later.

There are some other options on how to do this. We could do a double return value (Go style) where the second return value is the error message. Let me know what you think would be the best fit for the project.

patrickkwang: I think the use of `logger.error()` should be closely coupled with actually throwing errors and returning non-2xx status codes. My inclination is to move this logging step out to sit beside the `raise`.

alongreyber: The problem is that we don't know what error message to log if the logger.error call is in fetcher.py. The error could be "We couldn't find any possible plans starting from a pinned node" or "We couldn't find a KP for every edge in your query graph".

patrickkwang: Ah, I see the issue. To build on your point about separation of concerns, it should apply to logging, too. If `generate_plan()` is not going to raise an error, it shouldn't be logging one either. We could switch these logs to warnings, return an empty list, then log and raise an error when empty list becomes a terminal problem.

alongreyber: That makes sense

patrickkwang (approved):

PR #46, opened 2021-01-21
Predicate Directionality Fix

alongreyber:
Fixed issue where we were not traversing predicates correctly in the reverse direction. We do this by adding edges temporarily in the planning step to ensure that we cover plans that use edges in the opposite direction.

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -197,7 +197,7 @@ async def find_valid_permutations(
         # Swap subject and object
         reverse_edge['subject'], reverse_edge['object'] = \
             reverse_edge['object'], reverse_edge['subject']
-        reverse_edges[f"{edge_id}{REVERSE_EDGE_SUFFIX}"] = reverse_edge
+        reverse_edges[edge_id + REVERSE_EDGE_SUFFIX] = reverse_edge
 
         # Assign directionality to forward edge
         edge['predicate'] = [f"-{p}->" for p in edge['predicate']]
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -157,16 +157,7 @@ def expand_qg(
         new_predicate_list = []
         for t in edge['predicate']:
             new_predicate_list.extend(WBMT.get_descendants(t))
-
-        # For each edge predicate, we also need versions
-        # for -predicate-> and <-predicate-
-        # ltr = left to right
-        new_predicate_list_ltr = \
-            [f'-{p}->' for p in new_predicate_list]
-        new_predicate_list_rtl = \
-            [f'<-{p}-' for p in new_predicate_list]
-
-        edge['predicate'] = new_predicate_list_ltr + new_predicate_list_rtl
+        edge['predicate'] = new_predicate_list
 
     logger.debug({
         "description": "Expanded query graph with descendants",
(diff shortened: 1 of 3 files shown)


strider/query_planner.py
```diff
@@ -195,6 +188,23 @@ async def find_valid_permutations(
...
+    reverse_edges = {}
+    # Add reverse edges so that we can traverse edges in both directions
+    for edge_id, edge in expanded_qg['edges'].items():
+        reverse_edge = edge.copy()
+        reverse_edge['predicate'] = [
+            f"<-{p}-" for p in reverse_edge['predicate']]
+        # Swap subject and object
+        reverse_edge['subject'], reverse_edge['object'] = \
+            reverse_edge['object'], reverse_edge['subject']
+        reverse_edges[f"{edge_id}{REVERSE_EDGE_SUFFIX}"] = reverse_edge
```
patrickkwang, line 200: Admirable loyalty to f-strings.
```suggestion
        reverse_edges[edge_id + REVERSE_EDGE_SUFFIX] = reverse_edge
```

strider/query_planner.py
```diff
@@ -282,6 +293,10 @@ async def find_valid_permutations(
     for edge in expanded_qg['edges'].values():
         edge['predicate'] = list(edge.pop('filtered_predicates'))
 
+    # Remove edges with no predicates
```
patrickkwang, line 296: When does this happen?

strider/query_planner.py
```diff
@@ -390,8 +407,14 @@ async def generate_plans(
                 # Find edges that get us from
                 # There could be multiple edges that need to each
                 # be added as a separate step
-                step = Step(edge['subject'], edge_id, edge['object'])
                 op = get_operation(current_qg, edge)
+
+                # Reverse edges don't actually exist, so the steps in the plan
+                # should just refer to the forward edges
+                if edge_id.endswith(REVERSE_EDGE_SUFFIX):
+                    edge_id = edge_id.replace(REVERSE_EDGE_SUFFIX, '')
```
patrickkwang, line 415: Is the suffix just to prevent key collisions in `expanded_qg["edges"]`? It feels like there should be a neater way to do this...

patrickkwang (changes requested):

strider/query_planner.py
alongreyber: This happens on the two lines below

patrickkwang: Haha, sorry, I mean how would we encounter edges with no predicates? And why do we want to remove them?

strider/query_planner.py
alongreyber: Yeah it is. I can't really think of a neater way off the top of my head. We could store the reverse edges in a separate list, but that would be more messy. This way is nice because it lets us pretty easily handle the reverse edges when building the plan as well.

Right now this isn't bulletproof. Sending a query graph with a node key of 'e01-reversed' will break things. We could modify this to be a little better at handling that case, at the cost of code complexity.

strider/query_planner.py
alongreyber: Good question. This happens because of the filtering step where we remove all predicates where there are no KPs that offer solutions. So if we get a -biolink:regulates->, but we can only solve -biolink:positively_regulates->, we will make that replacement. 

If a predicate is filtered out entirely that means we can't solve this particular edge in this particular direction. That doesn't necessarily mean we have to throw out this query graph, because we might be able to solve the edge in the reverse direction.

alongreyber: The reason we have to filter it out is because the permute_qg function doesn't like being passed an empty array to permute (it will return nothing back).

patrickkwang (approved):

PR #47, opened 2021-01-21
Test Cleanup

alongreyber:
Switched to using the new utilities `query_graph_from_string` and `kps_from_string`. This makes the tests more readable because all of the information for most of the tests are contained in one place. In addition, this reduces the number of files in the repository. Builds on #46.


patrickkwang (approved):
This is very nice.

PR #56, opened 2021-01-25
Handle missing categories

patrickkwang:


strider/fetcher.py
```diff
@@ -116,11 +116,17 @@ async def setup(
...
         ))["query_graph"]
+
         # Fill in missing predicates with most general term
         for eid, edge in self.qgraph['edges'].items():
             if ('predicate' not in edge) or (edge['predicate'] is None):
                 edge['predicate'] = 'biolink:related_to'
 
+        # Fill in missing categories with most general term
+        for node in self.qgraph['nodes'].values():
+            if ('category' not in node) or (node['category'] is None):
```
alongreyber, line 127: This might be worth a separate issue, but we should be consistent with how we check for existence in a dictionary throughout the code. My preference is

```python
if not node.get('category', None):
```

because it covers both cases with a single condition. 

Alternatively, we could delete all None valued keys when we receive a query graph and then check for existence with the standard 
```python
'category' in node
```

I think this would be the best solution overall to keep the number of issues down to a minimum.

tests/server/test_server.py
```diff
@@ -113,6 +113,33 @@ async def test_solve_missing_predicate():
...
+@with_translator_overlay(
+    settings.kpregistry_url,
+    settings.normalizer_url,
+    [
+        ("ctd", CTD_PREFIXES),
+        ("hetio", DEFAULT_PREFIXES),
+        ("mychem", MYCHEM_PREFIXES),
+    ])
+async def test_solve_missing_category():
+    """Test solving the ex1 query graph, in which one of the categories is missing. """
```
alongreyber, line 126: For a test that covers a very specific condition like this, I think it's more readable to keep a simple query graph inside the test itself:
```
n0(( id MONDO:0005148 ))
n1(( category biolink:Drug ))
n0-- biolink:treated_by -->n1
```

alongreyber (approved):

strider/fetcher.py
patrickkwang: I agree, it might be best if this None == missing assumption only had to happen one place.

tests/server/test_server.py
patrickkwang: I agree. I was following the format that @cbizon used in the "missing predicate" test. At some point, we can go back and make both of these tests more readable.

PR #57, opened 2021-01-26
Fix node norm category handling

patrickkwang:


tests/query_planner/test_query_planner.py
```diff
@@ -324,3 +324,39 @@ async def testable_generate_plans():
...
+async def test_bad_norm():
+    """
+    Test that the pinned node "XXX:123" that the normalizer does not know
+    is still handled correctly based on the provided category.
+    """
+
+    qg = {
+        "nodes": {
+            "n0": {
+                "id": "XXX:123",
```
alongreyber, line 345: Could you use the new query graph string utility for consistency?

tests/query_planner/test_query_planner.py
```diff
@@ -324,3 +324,39 @@ async def testable_generate_plans():
         testable_generate_plans,
         "generate plan for a typical query graph (50k KPs)",
     )
+
+
+@pytest.mark.asyncio
```
alongreyber, line 329: I am trying to keep the tests marked with @pytest.longrun at the bottom of this file. Can you move this test above those?

strider/normalizer.py
```diff
@@ -13,6 +16,12 @@ async def get_types(self, curies):
                 params=dict(curie=curies)
             )
         types = []
+        if response.status_code >= 300:
+            LOGGER.warning(
+                f"Received {response.status_code} response from normalizer: "
+                + response.text
+            )
+            return types
         for c in curies:
```
alongreyber, line 25: Maybe we should add a warning to the log if we don't get any results back from normalizer. If we do, we should also add a check to the test to make sure the log message exists.

alongreyber (changes requested):

tests/query_planner/test_query_planner.py
patrickkwang: I don't think so... The query graph string function appears to take only a single property per node, we need two here. We could probably improve it.

strider/normalizer.py
patrickkwang: You mean also warn if we get an empty or null 200 response? Yes, that's a good idea.

tests/query_planner/test_query_planner.py
alongreyber: Ah good point. I didn't notice that.

alongreyber (changes requested):
If you rebase this branch onto master and then run the tests, `test_invalid_two_disconnected_components` fails because of an invalid log message. I suspect that this is because MONDO:0005737 is not available in the normalizer. Could you add a fix?

patrickkwang:
Good catch, @alongreyber. If we want, we can enforce that PRs should be up-to-date with main. It leads to more force-pushing than necessary, though.

alongreyber (approved):

PR #58, opened 2021-01-26
Verify log contents in tests

alongreyber:
Added utilities for verifying log contents in each test to ensure we fail tests if there are unexpected warnings or errors. 

(This should catch an issue when combined with #57 😄)


patrickkwang (approved):
Very cool!

PR #59, opened 2021-01-26
Management Script

alongreyber:
Added management script and updated README with instructions on how to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ A live version of the API can be found [here](http://robokop.renci.org:5781/docs
 The codebase comes with a zero-dependency python management script that can be used to automate basic local development tasks. Make sure you have docker and docker-compose installed and then run:
 
 ```bash
-python3 manage.py dev # starts server accessible at 5781
-python3 manage.py test # run tests
-python3 manage.py coverage # run tests, open coverage report in browser
+./manage.py dev # starts server accessible at 5781
+./manage.py test # run tests
+./manage.py coverage # run tests, open coverage report in browser
 ```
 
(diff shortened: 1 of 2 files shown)


patrickkwang (commented):
Is there a reason not to have a `prod` argument, too?

alongreyber:
Generally I don't have a prod argument in these types of scripts because running a production environment on a local environment might not work (for example, if a required service is only accessible from a deployment network). In this case though, we can run production locally, so I can add it if you would like.

README.md
```diff
@@ -11,23 +11,28 @@ A live version of the API can be found [here](http://robokop.renci.org:5781/docs
...
 ## Local Development
 
-Docker-compose can be used to quickly set up a locally running version of the app. Once you have installed docker-compose you can run:
+### Management Script
+
+The codebase comes with a zero-dependency python management script that can be used to automate basic local development tasks. Make sure you have docker and docker-compose installed and then run:
 
 ```bash
-docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build
+python3 manage.py dev # starts server accessible at 5781
```
patrickkwang, line 19: Python 2 is officially dead, so I think it's safe to just use `python`. Alternatively, we could make `manage.py` executable with `#!/usr/bin/env python`.

alongreyber: Python 2.7 is still installed on Mac OS by default (it's the default on my system) so leaving this as python3 would be convenient for me. I do think making the script executable would be good, I'll add that.

patrickkwang (approved):

PR #73, opened 2021-02-01
List fixes

### LINKED ISSUE #60: Query Internal Server Error
The following query causes an internal server error:
```json
{
  "message": {
    "query_graph": {
      "nodes": {
        "n0": {
          "category": "biolink:Disease",
          "id": [
            "MONDO:0005737"
          ]
        },
        "n1": {
          "category": "biolink:Gene"
        },
        "n2": {
          "category": "biolink:Disease"
        },
        "n3": {
          "category": "biolink:ChemicalSubstance"
        }
      },
      "edges": {
        "e0": {
          "subject": "n0",
          "object": "n1"
        },
        "e1": {
          "subject": "n1",
          "object": "n2"
        },
        "e2": {
          "subject": "n1",
          "object": "n3"
        },
        "e3": {
          "subject": "n3",
          "object": "n2",
          "predicate": "biolink:treats"
        }
      }
    }
  }
}
```

alongreyber:
Allows query graphs with list properties (category, id, predicate). Fixes #60.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -5,6 +5,6 @@ pytest==5.4.3
 pytest-asyncio==0.14.0
 pytest-cov==2.10.0
 git+git://github.com/TranslatorIIPrototypes/kp_registry@v1.0
-git+git://github.com/ranking-agent/simple-kp@master
+git+git://github.com/ranking-agent/simple-kp@v1.0.1
 git+git://github.com/patrickkwang/small-kg@main
 git+git://github.com/patrickkwang/asgiar@main
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -5,6 +5,6 @@ pytest==5.4.3
 pytest-asyncio==0.14.0
 pytest-cov==2.10.0
 git+git://github.com/TranslatorIIPrototypes/kp_registry@v1.0
-git+git://github.com/ranking-agent/simple-kp@v1.0
+git+git://github.com/ranking-agent/simple-kp@master
 git+git://github.com/patrickkwang/small-kg@main
 git+git://github.com/patrickkwang/asgiar@main
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -230,7 +230,10 @@ class StriderWorker(Worker):
 
         response = await self.execute_step(
             step,
-            result["node_bindings"][step.source][0]["id"],
+            # Use a frozen set because it is hashable
+            frozenset(
+                result["node_bindings"][step.source][0]["id"]
+            ),
         )
 
         # process kgraph
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -150,18 +150,18 @@ class StriderWorker(Worker):
 
         self.logger.debug({"plan": self.plan})
 
-        # add first partial result
+        # add partial results for each curie that we are given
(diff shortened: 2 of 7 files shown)


strider/trapi.py
```diff
@@ -58,11 +58,10 @@ def get_curies(message: Message) -> list[str]:
...
     curies = set()
     if 'query_graph' in message:
-        curies |= {
-            qnode.get("id")
-            for qnode in message['query_graph']['nodes'].values()
-            if qnode.get("id", False)
-        }
+        for qnode in message['query_graph']['nodes'].values():
+            qnode_id = qnode.get("id", False)
+            if qnode_id:
```
patrickkwang, line 63: This is a good example use case for the walrus operator.

strider/trapi.py
```diff
@@ -58,11 +58,10 @@ def get_curies(message: Message) -> list[str]:
...
+                curies.update(qnode_id)
```
patrickkwang, line 64: I like the new `|=` for this; it's clearer to me.

strider/fetcher.py
```diff
@@ -150,18 +150,18 @@ async def generate_plan(self):
...
                 continue
-            result = {
-                "node_bindings": {
-                    qnode_id: [{"id": qnode["id"]}]
-                },
-                "edge_bindings": {},
-            }
-            await self.put(result)
-            break
+            for curie in qnode["id"]:
```
patrickkwang, line 157: Do we have a test for this?

requirements-test.txt
```diff
@@ -5,6 +5,6 @@ pytest==5.4.3
 pytest-asyncio==0.14.0
 pytest-cov==2.10.0
 git+git://github.com/TranslatorIIPrototypes/kp_registry@v1.0
-git+git://github.com/ranking-agent/simple-kp@v1.0
+git+git://github.com/ranking-agent/simple-kp@master
```
patrickkwang, line 8: I believe we can use the `v1.0.0` tag here.

patrickkwang (changes requested):
Very nice. Minor comments to be addressed.

strider/trapi.py
alongreyber: How should this use the walrus operator? Do you mean something like this:
```python
qnode_id |= qnode.get("id", None)
```

strider/fetcher.py
alongreyber: What do you mean exactly? This line has coverage from a couple of the tests in tests/server/test_server.py. Is there a specific condition you had in mind to test?

strider/trapi.py
patrickkwang: `if qnode_id := qnode.get("id", False):` I think.
https://www.python.org/dev/peps/pep-0572/

strider/fetcher.py
patrickkwang: I mean a test for a qnode with multiple `id`s. Probably it's tied in with the multiple-plan issues.

alongreyber: If I make one, it will be failing. Should I still add it to this commit, or wait for your multiple plan issues?

patrickkwang: That's okay, we'll add a test later.

patrickkwang (approved):

PR #83, opened 2021-02-03
Fix mutability bug

patrickkwang:


tests/server/test_server.py
```diff
@@ -402,3 +402,57 @@ async def test_subpredicate():
...
+
+    # Create query
+    q = Query(
+        message=Message(
+            query_graph=QueryGraph.parse_obj(qg)
+        )
+    )
+
+    # Run
+    output = await sync_query(q)
```
alongreyber, line 457: What does this test exactly? It looks like we're just checking if this query works properly. Would this fail without the mutability fix?

alongreyber (approved):

tests/server/test_server.py
patrickkwang: It does, yes.

PR #84, opened 2021-02-03
Fixed plan endpoint

alongreyber:
Plan endpoint was trying to use a tuple as a JSON key and was therefore throwing a 500 error. This replaces the tuple with a simple string key format (`n0-e01->n1`).

### DIFF
--- a/strider/server.py
+++ b/strider/server.py
@@ -370,7 +370,7 @@ async def generate_traversal_plan(
     for plan in plans:
         plans_stringified_keys.append(
             {
-                f"{step.source}-{step.edge}->{step.target}": v
+                f"{step.source}-{step.edge}-{step.target}": v
                 for step, v in plan.items()
             }
         )
--- a/strider/server.py
+++ b/strider/server.py
@@ -8,7 +8,7 @@ import os
 import enum
 from pathlib import Path
 import pprint
-from typing import Dict
+from typing import List, Dict
 
 from fastapi import Body, Depends, FastAPI, HTTPException, BackgroundTasks
 from fastapi.openapi.docs import (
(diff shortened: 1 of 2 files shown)


strider/server.py
```diff
@@ -359,13 +359,23 @@ async def extract_results(query_id, since, limit, offset, database):
     ]
 
 
-@APP.post('/plan', response_model=Dict)
+@APP.post('/plan', response_model=List[Dict])
```
patrickkwang, line 362: Elsewhere we use `list` and `dict` as allowed by 3.9.

strider/server.py
```diff
@@ -359,13 +359,23 @@ async def extract_results(query_id, since, limit, offset, database):
...
     """Generate plans for traversing knowledge providers."""
     query_graph = query.message.query_graph.dict()
-    return await generate_plans(query_graph)
+    plans = await generate_plans(query_graph)
+
+    plans_stringified_keys = []
+    for plan in plans:
+        plans_stringified_keys.append(
+            {
+                f"{step.source}-{step.edge}->{step.target}": v
```
patrickkwang, line 374: Doesn't the arrow direction depend on the query graph?

patrickkwang (commented):
A couple questions.

strider/server.py
alongreyber: In this case the arrow indicates the source and target of the edge in the query graph. The predicate direction is indicated in the plan itself. I can remove the arrow and just have n0-e01-n1 if that would be less confusing.

patrickkwang: Our convention is that the operation direction is indicated by the order of the nodes on the line and in most (all?) contexts, an arrow indicates predicate direction, so I think leaving it off here is a good choice.

patrickkwang (approved):

PR #89, opened 2021-02-08
Result deduplication

alongreyber:
Added a test to ensure that if we receive duplicate results from a KP the results will not be added multiple times to the resulting message. De-duplication is implemented using the Pydantic Result model.

### DIFF
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -27,6 +27,7 @@ from .caching import async_locking_cache
 from .storage import RedisGraph, RedisList, RedisLogHandler
 from .kp_registry import Registry
 from .config import settings
+from .util import ensure_list
 
 # Initialize registry
 registry = Registry(settings.kpregistry_url)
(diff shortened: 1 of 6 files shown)


strider/fetcher.py
```diff
@@ -253,6 +258,8 @@ def get_kp_request_body(
         curie: str,
         step: Step,
         kp: dict,
+
+
```
patrickkwang, line 262: What's the deal with these extra newlines?

alongreyber: Not sure, but I removed them.

strider/trapi.py
```diff
@@ -15,14 +15,18 @@ def merge_messages(messages: list[Message]) -> Message:
     for message in messages:
         knodes |= message["knowledge_graph"]["nodes"]
         kedges |= message["knowledge_graph"]["edges"]
-        results.extend(message["results"])
+        for result in message["results"]:
+            # Use pydantic model so that we can deduplicate
+            result_pydantic = Result.parse_obj(result)
```
patrickkwang, line 20: This is handy, but also encroaches on a nice separation we have between the web server that handles pydantic objects, and the inner methods, which don't. The other risk(?) here is that the results actually don't validate against the Result model, but in a way that Strider is OK with. We are deliberately lax in enforcing TRAPI compliance because we want to get away with using slightly naughty KPs if we can.

alongreyber: In that case, how do we want to compare result objects? Do they need to be exactly equal, including extra fields? Or do we only care about specific fields, and if so which ones?

patrickkwang: I think we want to ignore the binding properties except for `id` and make sure list order doesn't matter.
```
{
    key: {bound["id"] for bound in value}
    for key, value in result["node_bindings"].items()
}
```

patrickkwang (approved):
Very nice.

PR #91, opened 2021-02-10
Normalizer Overlay Changes

alongreyber:
Added ability to specify normalizer data using a string format.

### DIFF
--- a/tests/helpers/utils.py
+++ b/tests/helpers/utils.py
@@ -125,7 +125,7 @@ def kps_from_string(s):
     return kps
 
 
-def node_normalizer_data_from_string(s):
+def normalizer_data_from_string(s):
     """
     Parse data for node normalizer from string. Useful for tests.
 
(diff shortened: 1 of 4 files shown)


tests/helpers/utils.py
```diff
@@ -125,6 +125,42 @@ def kps_from_string(s):
...
+    # so we use inspect.cleandoc to remove leading indentation
+    s = inspect.cleandoc(s)
+
+    category_mappings = {}
+    synset_mappings = {}
+    for line in s.splitlines():
+        tokens = line.split(" ")
+
+        curie = tokens[0]
+        if curie not in category_mappings:
```
patrickkwang, line 148: Should we use `defaultdict(list)` instead?

patrickkwang (changes requested):
Looks great! One minor suggestion.

patrickkwang (approved):

PR #93, opened 2021-02-10
Category Normalizer Mismatch

### LINKED ISSUE #62: Query plan not contacting all available KPs
The following query graph is not contacting a KP that advertises AnatomicalEntity.
```json
{
  "message": {
    "query_graph": {
      "nodes": {
        "n0": {
          "category": "biolink:Gene"
        },
        "n1": {
          "category": "biolink:AnatomicalEntity",
          "id": "GO:0005634"
        }
      },
      "edges": {
        "e0": {
          "subject": "n0",
          "object": "n1",
          "predicate": "biolink:part_of"
        }
      }
    }
  }
}
```

alongreyber:
Fixed #62 where the category given does not match the one that is in the normalizer. The fix involves passing the expanded query graph into StriderWorker and then using that for partial results.

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -7,11 +7,8 @@ from typing import Generator
 from reasoner_pydantic import QueryGraph
 
 from strider.kp_registry import Registry
-from strider.normalizer import Normalizer
 from strider.util import WrappedBMT
 from strider.config import settings
-
-WBMT = WrappedBMT()
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


patrickkwang (commented):
What if we move the qgraph expansion out of `generate_plans()`? It seems to me that we can basically do it as soon as we receive the qgraph, and throw away the original. This isn't really like an intermediate artifact, it's just how we're interpreting the qgraph. Would that cause problems?

alongreyber:
I'm not sure if it will cause problems, but I will give it a try. It would make things a bit cleaner.

strider/server.py
```diff
@@ -358,7 +358,7 @@ async def generate_traversal_plan(
 ) -> list[dict]:
     """Generate plans for traversing knowledge providers."""
     query_graph = query.message.query_graph.dict()
-    plans = await generate_plans(query_graph)
+    plans, _ = await generate_plans(query_graph)
```
patrickkwang, line 361: I suspect this is broken. We should write tests for the actual endpoints using FastAPI's TestClient.

strider/query_planner.py
```diff
@@ -391,18 +296,19 @@ def dfs(operation_graph, source) -> (list[str], list[str]):
 async def generate_plans(
         qgraph: QueryGraph,
         kp_registry: Registry = None,
-        normalizer: Normalizer = None,
         logger: logging.Logger = LOGGER,
-) -> list[dict[Step, list]]:
+) -> (list[dict[Step, list]], dict[str, dict]):
```
patrickkwang, line 300: This output signature needs to revert.

tests/server/test_server.py
```diff
@@ -213,6 +213,50 @@ async def test_solve_missing_category():
...
+        """
+            CHEBI:6801(( category biolink:Drug ))
+            MONDO:0005148(( category biolink:Disease ))
+            CHEBI:6801-- predicate biolink:treats -->MONDO:0005148
+        """
+    },
+    normalizer_data="""
+        CHEBI:6801 categories biolink:Drug
+        """
+
```
patrickkwang, line 231: Extra newline?

patrickkwang (changes requested):
I like this organization much better! Minor fixes.

patrickkwang (approved):

PR #95, opened 2021-02-12
Loop Support

alongreyber:
Modified planning algorithm to handle loop support. Added tests for both planning and solving loops.

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -259,7 +259,7 @@ def dfs(operation_graph, source) -> (list[str], list[str]):
 
     Returns a list of nodes and edges
     that we traversed. Will not go back to
-    previously visited nodes or edges.
+    previously visited edges.
     """
     path_nodes = [source]
     path_edges = []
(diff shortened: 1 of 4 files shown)


strider/util.py
```diff
@@ -243,3 +243,7 @@ def remove_null_values(obj):
         ]
     else:
         return obj
+
+
+def last(alist, n):
+    return alist[:-n] if n else alist[:]
```
patrickkwang, line 249: This is confusing me. You're not getting the last `n`, you're getting the first `len(alist) - n`. Is that the intent?

strider/query_planner.py
```diff
@@ -253,44 +253,94 @@ async def find_valid_permutations(
...
-    # The most efficient way to do a depth first search
-    # is with an adjacency list, so we build one here
+    # Initialize stack with all edges that start at our source node
+    stack = []
+    for edge in starting_edges:
+        new_graph = copy.deepcopy(operation_graph)
+        del new_graph['edges'][edge]
+        stack.append((new_graph, [edge]))
 
+    # Recursively finish paths
```
patrickkwang, line 276: Where does it recurse?

strider/query_planner.py
```diff
@@ -253,44 +253,94 @@ async def find_valid_permutations(
...
+        current_node = operation_graph['edges'][current_path[-1]]['target']
+
+        adjacent_edges = [
+            edge_id for edge_id, edge in current_graph['edges'].items()
+            if edge['source'] == current_node
+        ]
+
+        if len(adjacent_edges) == 0:
+            # Done
+            paths.append(current_path)
```
patrickkwang, line 292: Is this guaranteed to be an Eulerian path? This seems to check only that there's nowhere else we can go from here, not that there are no edges left to be traversed.

alongreyber:
I am working on an updated planning method based on edge reification, but this commit might be able to help with loop queries in the meantime.

patrickkwang (approved):

PR #97, opened 2021-02-17
Reified Query Planning

alongreyber:
Switched to a planning using a reified graph where all query graph nodes and edges are represented as nodes.

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -180,27 +180,36 @@ async def qg_to_og(
     return ograph
 
 
-async def filter_categories_predicates(operation_graph, operation_kp_map):
+async def filter_categories_predicates(graph, operation_kp_map):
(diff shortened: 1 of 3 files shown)


strider/query_planner.py
```diff
@@ -169,179 +182,116 @@ async def qg_to_og(
...
-async def find_valid_permutations(
-        operation_graph,
-        kp_registry: Registry = None,
-        logger: logging.Logger = LOGGER,
-) -> list[dict]:
-    """
-    Given an operation graph, generate a list of operation graphs
-    that are solvable, and annotate those with a 'request_kp' property
-    that shows which KPs to contact to get results.
+def get_next_nodes(graph, path):
```
patrickkwang, line 233: Type hints might help here, especially because this graph format is not the typical one we deal with.

patrickkwang (approved):

PR #106, opened 2021-03-08
Fixed double-sided predicate

alongreyber:
Fixed issue where double sided predicate adds a KP four times to the plan. Added a test and a fix.

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -276,18 +276,23 @@ async def filter_categories_predicates(query_graph, operation_graph):
             if edge["qg_edge_id"] == qg_edge_id
         ]
 
-        for og_edge in associated_operation_graph_edges:
+        sub = query_graph['nodes'][edge['subject']]
+        obj = query_graph['nodes'][edge['object']]
 
-            sub = query_graph['nodes'][edge['subject']]
-            obj = query_graph['nodes'][edge['object']]
-            if og_edge["qg_traversal_reverse"]:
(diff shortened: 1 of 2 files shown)


strider/query_planner.py
```diff
@@ -270,15 +270,19 @@ async def filter_categories_predicates(query_graph, operation_graph):
...
             edge for edge in operation_graph["edges"].values()
             if edge["qg_edge_id"] == qg_edge_id
         ]
 
         for og_edge in associated_operation_graph_edges:
+
+            sub = query_graph['nodes'][edge['subject']]
+            obj = query_graph['nodes'][edge['object']]
+            if og_edge["qg_traversal_reverse"]:
+                sub, obj = obj, sub
```
patrickkwang, line 284: Should these be named "source" and "target" instead?

alongreyber: I don't really know. They refer to query graph nodes, not operation graph nodes, so I thought subject and object would be more appropriate.

patrickkwang: It looks to me like the goal for these is to represent the source and target of an operation, below when we add stuff called "source_category" and "target_category". Are those named appropriately? If I'm right, here's what I would do. Pull lines 281 and 282 out of the loop. Make the `if` block inside an `if...else` that declares `source` and `target` appropriately, and use those below.

patrickkwang (approved):

PR #108, opened 2021-03-12
Descendant Reverse Fix

alongreyber:
Added a test for a case where the query graph includes related_to and we need to subclass related_to in both forward and reverse directions. Implemented this change by switching query graph permutations to operation graph permutations.

### DIFF
--- a/tests/query_planner/test_query_planner.py
+++ b/tests/query_planner/test_query_planner.py
@@ -1,13 +1,9 @@
 from pathlib import Path
 from functools import partial
-import os
-import math
 import json
 import logging
-import random
 
 import pytest
-from bmt import Toolkit as BMToolkit
 
 from tests.helpers.context import \
     with_registry_overlay, with_norm_overlay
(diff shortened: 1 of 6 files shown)


tests/query_planner/ex1_qg.json
```diff
@@ -1,25 +1,25 @@
...
-      "object" : "n0",
-      "predicate" : "biolink:treated_by"
-    },
-    "e02" : {
-      "subject" : "n0",
-      "object" : "n2",
-      "predicate" : "biolink:affects_abundance_of"
-    }
-  }
+	"nodes": {
```
patrickkwang, line 2: Are these tabs?

tests/query_planner/test_query_planner.py
```diff
@@ -412,7 +421,8 @@ async def test_unbound_unconnected_node(caplog):
         n2(( category biolink:PhenotypicFeature ))
         """
     )
-    qg = await expand_qg(qg, logging.getLogger())
+    await fill_categories_predicates(qg, logging.getLogger())
```
patrickkwang, line 424: We seem to do these together a lot. Should we make a wrapper for these two functions?

tests/query_planner/ex1_qg.json
alongreyber: Not as far as I can tell

alongreyber: Okay they actually are tabs - I'm working on a fix

patrickkwang (approved):

PR #109, opened 2021-03-12
Added functions to help with validating plans

alongreyber:
Added helper functions for validating the contents of plans in a more strict way. Right now, these functions are only used for one test (to demonstrate their usage) but if we like this structure I can go through and convert the rest of the tests to use this.


alongreyber:
This change caught a bug related to predicate filtering (hence the failing test) but it's going to be easier to fix once #108 is merged in so I'm changing this to draft for now.

patrickkwang (approved):

PR #115, opened 2021-03-23
Switched permutations to use KPs

alongreyber:
Switched query planning to use KPs for permutations. Also added an optimization to not use copy.deepcopy during the permutations.

### DIFF
--- a/strider/query_planner.py
+++ b/strider/query_planner.py
@@ -70,7 +70,7 @@ def permute_graph(graph: dict) -> Generator[dict, None, None]:
             continue
 
         # Any fields on an edge that might need to be permuted
-        edge_fields_to_permute = ['forward_kps', 'reverse_kps']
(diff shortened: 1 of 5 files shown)


strider/query_planner.py
```diff
@@ -240,6 +285,50 @@ async def qg_to_og(qgraph):
...
 
+def fix_categories_predicates(query_graph):
+    """
+    Given a permuted query graph with one KP,
+    fix the node categories and predicates to match the KP
+
+    Returns false if there is a conflicting node type
+    """
+    for edge in query_graph['edges'].values():
+        kp = edge["kps"]
```
patrickkwang, line 296: Is edge["kps"] sometimes a list and sometimes a scalar? Looks like this comes from `permute_graph()`. It might be more readable if `permute_graph()` produced length-1 lists rather than scalars.

Issue #102, opened 2021-03-02
Not returning Automat KEGG info?


cbizon:
I am trying to establish a connection between the gene GK and the chemical glycerol.  We have an edge:
(GK)-[increases degradation of]->(glycerol) in KEGG automat, but I can't seem to get it from strider.

If I give this query:
```
{
    "message": {
        "query_graph": {
            "nodes": {
                "a": {
                    "category": "biolink:ChemicalSubstance"
                },
                "b": {
                    "category": "biolink:Gene",
                    "id": "UniProtKB:P32189"
                }
            },
            "edges": {
                "ab": {
                    "subject": "a",
                    "object": "b"
                }
            }
        }
    },
    "knowledge_graph": {
        "nodes": [],
        "edges": []
    },
    "results": []
}
```

I don't get any results.  But if I change gene ID to the NCBIGene identifier instead, I do get results.  It looks like nodenorm handles this id fine, so don't know why that would be.

However, if I use the NCBIGene, I get many chemicals, including gycerol, but the edge to glycerol is an interacts_with edge from hmdb automat. That's true, but I don't see the degrades edge from KEGG.

(Came about as a result of 3/1 mini hackathon)

cbizon:
Note: now with this query as written above, I am getting some results.  However, I'm still not getting the KEGG result I expect.  Also, I am seeing this kind of thing in the results:
```
{'node_bindings': {'b': [{'id': 'UniProtKB:P32189',
       'category': 'biolink:Gene'},
      {'id': 'NCBIGENE:2710', 'category': 'biolink:Gene'}],
     'a': [{'id': 'MESH:D000517', 'category': 'biolink:ChemicalSubstance'}]},
    'edge_bindings': {'ab': [{'id': 'MESH:D000517-NCBIGENE:2710-SEMMED Gene API-SEMMED'}]}},
```
For the 'b' node, which is defined in the query graph with a particular id, we are binding to 2 KG nodes, which should be the same node becuase Uniprotkb:P32189 == NCBIGENE:2710.

As an uposhot ( I suspect ) the GammaViewer also gets confused and shows the gene as a missing node:
![image](https://user-images.githubusercontent.com/10744350/111669223-f48f5d80-87ec-11eb-8310-51362f226f69.png)

cbizon:
Using NCBIGENE in the initial query doesn't produce the doubled node, and GammaViewer works slightly better, but I still am not getting the KEGG results.

alongreyber:
[...]

PR #119, opened 2021-04-01
Issue #102 Fix

alongreyber:
Fixed issues with the query in #102. This includes the following changes:

  - Switched from category and predicate permutations to KP permutations
  - Message output verification in some tests
  - Permutation code optimization by removing `copy.deepcopy`

### DIFF
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -28,7 +28,7 @@ from .caching import async_locking_cache
 from .storage import RedisGraph, RedisList, RedisLogHandler
 from .kp_registry import Registry
 from .config import settings
-from .util import ensure_list, standardize_graph_lists
+from .util import ensure_list, standardize_graph_lists, extract_predicate_direction
 
 # Initialize registry
 registry = Registry(settings.kpregistry_url)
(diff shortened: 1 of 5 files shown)


strider/fetcher.py
```diff
@@ -257,26 +263,34 @@ def get_kp_request_body(
...
+    predicate, _ = \
+        extract_predicate_direction(kp["edge_predicate"])
+
+    # Build request edges
+    request_edge = qgraph["edges"][step.edge].copy()
+    request_source = qgraph["nodes"][step.source].copy()
+    request_target = qgraph["nodes"][step.target].copy()
+
+    # Update request properties to match what the KP expects
+    request_edge["subject"] = step.source
```
patrickkwang, line 276: It makes me nervous any time we have "subject" and "source", or "object" and "target", in the same line. The logic here is not obvious.

strider/query_planner.py
```diff
@@ -225,58 +291,61 @@ async def qg_to_og(qgraph):
...
 
 
-def filter_categories_predicates(operation_graph):
+def fix_categories_predicates(query_graph):
     """
-    Filter out categories and predicates that are not defined in any KP
+    Given a permuted query graph with one KP,
+    fix the node categories and predicates to match the KP
+
+    Returns false if there is a conflicting node type
```
patrickkwang, line 314: It seems probably more "pythonic" to raise an exception for conflicting categories rather than returning a boolean.

tests/server/test_server.py
```diff
@@ -654,8 +680,24 @@ async def test_inverse_predicate():
...
+        "results": [
+            """
+            node_bindings:
+                n0 MONDO:0005148
+                n1 CHEBI:6801
+            edge_bindings:
+                n0n1 CHEBI:6801-MONDO:0005148
+            """
+        ]
+    },
```
patrickkwang, line 699: Whitespace is weird here.

alongreyber: How do you think this whitespace should look? I think it's kinda ugly right now, but I'm not exactly sure how to make it better.

patrickkwang (approved):

PR #127, opened 2021-04-07
Synonym & Prefix Fixes

alongreyber:
Fixed bug with synonyms, and switched prefixes to come from BMT


strider/util.py
```diff
@@ -15,19 +15,42 @@ def snake_to_camel(s):
...
+            value = f(lookup)
+            if value is None:
+                raise KeyError
+            else:
+                return value
+
+        def __contains__(self, lookup):
+            return f(lookup) is not None
+
+    map_instance = Mapping()
```
patrickkwang, line 34: I'd just `return Mapping()`.

tests/helpers/utils.py
```diff
@@ -140,12 +140,18 @@ def normalizer_data_from_string(s):
...
-    category_mappings = defaultdict(list)
-    synset_mappings = defaultdict(list)
+    category_mappings = {}
+    synset_mappings = {}
     for line in s.splitlines():
         tokens = line.split(" ")
 
         curie = tokens[0]
+        if curie not in category_mappings:
+            category_mappings[curie] = []
```
patrickkwang, line 150: This is exactly what `defaultdict(list)` does, right? Why the change?

tests/helpers/utils.py
```diff
@@ -140,12 +140,18 @@ def normalizer_data_from_string(s):
...
+        if curie not in synset_mappings:
+            # Curies are always self-synonyms
+            synset_mappings[curie] = [curie]
```
patrickkwang, line 153: Is line 160 now unreachable? Why do this instead of `defaultdict`? What happens if `curie` _is_ in `synset_mappings`? Is that possible, and would it then not be its own synonym?

strider/util.py
```diff
@@ -87,6 +110,17 @@ def predicate_inverse(self, predicate):
             return self.old_case_to_new_case(predicate_inverse_old_format)
         return None
 
+    def entity_prefixes(self, entity):
+        """ Get prefixes for a given entity """
+        old_format = self.new_case_to_old_case(entity)
+        element = self.bmt.get_element(old_format)
```
patrickkwang, line 116: How fast is this? We could memoize this function if it's helpful. Then we'd have a map pretending to be a function pretending to be a map!

patrickkwang (changes requested):

tests/helpers/utils.py
alongreyber: Our code relies on the normalizer returning the given curie in the list of equivalent_identifiers. I confirmed that this is the case in the real node normalizer:

Screen Shot 2021-04-07 at 1 33 33 PM

This was the bug that caused test_issue_102 to fail. If the curie is not listed in equivalent_identifiers we never do any synonymizing. You're correct that we can use a defaultdict to streamline this and that line 160 is unreachable.

patrickkwang (approved):

PR #129, opened 2021-04-09
Added conversion of Protein to GeneOrGeneProduct

### LINKED ISSUE #124: Respond to queries with biolink:Protein
See discussion here: 
https://github.com/NCATSTranslator/testing/issues/38

biolink has Gene and Protein categories.  We treat these as the same thing.  (there are good reasons to do this and other good reasons not to, but right now we are).

But, we are calling the merged category Gene, so if somebody queries with Protein, we are not returning data that is relevant.  We should internally replace any biolink:Protein node with a biolink:Gene node and run that query through strider.

alongreyber:
Also added test. Fixes #124.

### DIFF
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -131,9 +131,9 @@ class StriderWorker(Worker):
         # Replace biolink:Proten with biolink:GeneOrGeneProduct
         for node in self.qgraph["nodes"].values():
             categories = node.get("category", [])
-            for category in categories:
+            for index, category in enumerate(categories):
                 if category == "biolink:Protein":
-                    category = "biolink:GeneOrGeneProduct"
+                    categories[index] = "biolink:GeneOrGeneProduct"
 
     async def generate_plan(self):
         """
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -130,10 +130,13 @@ class StriderWorker(Worker):
 
         # Replace biolink:Protein with biolink:GeneOrGeneProduct
         for node in self.qgraph["nodes"].values():
-            categories = node.get("category", [])
-            for index, category in enumerate(categories):
-                if category == "biolink:Protein":
-                    categories[index] = "biolink:GeneOrGeneProduct"
+            if not node.get("category"):
+                continue
+            node["category"] = [
+                "biolink:GeneOrGeneProduct"
+                if category == "biolink:Protein" else category
+                for category in node["category"]
+            ]
 
     async def generate_plan(self):
         """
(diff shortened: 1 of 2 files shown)


strider/fetcher.py
```diff
@@ -124,10 +124,17 @@ async def setup(
             self.preferred_prefixes,
         ))["query_graph"]
 
-        # Expand the query graph using the biolink model hierarchies
+        # Fill in missing categories and predicates using normalizer
         await fill_categories_predicates(self.qgraph, self.logger)
         standardize_graph_lists(self.qgraph)
 
+        # Replace biolink:Proten with biolink:GeneOrGeneProduct
```
patrickkwang, line 131:
```suggestion
        # Replace biolink:Protein with biolink:GeneOrGeneProduct
```

tests/server/test_server.py
```diff
@@ -1008,3 +1008,44 @@ async def test_solve_multiple_reverse_edges():
...
+
+    # Create query
+    q = Query(
+        message=Message(
+            query_graph=QueryGraph.parse_obj(QGRAPH)
+        )
+    )
+
+    # Run
+    output = await sync_query(q)
```
patrickkwang, line 1051: Do we also need to test that the results are not empty?

strider/fetcher.py
```diff
@@ -124,10 +124,17 @@ async def setup(
...
+        for node in self.qgraph["nodes"].values():
+            categories = node.get("category", [])
+            for category in categories:
+                if category == "biolink:Protein":
+                    category = "biolink:GeneOrGeneProduct"
```
patrickkwang, line 136: I suspect that this does not actually edit the qgraph and that the test is not testing what you want it to test.

patrickkwang (changes requested):

strider/fetcher.py
```diff
@@ -124,10 +124,17 @@ async def setup(
...
         ))["query_graph"]
 
-        # Expand the query graph using the biolink model hierarchies
+        # Fill in missing categories and predicates using normalizer
         await fill_categories_predicates(self.qgraph, self.logger)
         standardize_graph_lists(self.qgraph)
 
+        # Replace biolink:Protein with biolink:GeneOrGeneProduct
+        for node in self.qgraph["nodes"].values():
+            categories = node.get("category", [])
```
patrickkwang, line 133: I think this can be done more cleanly. Something like this:
```python
if not node.get("category"):
    continue
node["category"] = [
    "biolink:GeneOrGeneProduct" if category == "biolink:Protein" else category
    for category in node["category"]
]
```

patrickkwang (approved):

PR #131, opened 2021-04-09
Local run tweaks

alongreyber:
Added a couple small changes to make it easier to debug a query with pdb:

- Enabled python-dotenv
- Added run.py file

### DIFF
--- a/run.py
+++ b/run.py
@@ -1,0 +1,7 @@
+import uvicorn
+
+if __name__ == "__main__":
+    uvicorn.run("strider.server:APP",
+                host="0.0.0.0",
+                port=5781,
+                reload=True)
--- a/strider/storage.py
+++ b/strider/storage.py
@@ -6,7 +6,7 @@ import logging
 
 from .config import settings
 
-if settings.redis_url.startswith('redis://fakeredis'):
+if settings.redis_url == "redis://fakeredis:6379/0":
     import fakeredis
     r = fakeredis.FakeRedis(
         encoding="utf-8",
--- a/strider/storage.py
+++ b/strider/storage.py
@@ -6,7 +6,7 @@ import logging
 
 from .config import settings
 
-if settings.redis_url == 'redis://fakeredis':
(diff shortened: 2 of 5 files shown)


strider/storage.py
```diff
@@ -6,7 +6,7 @@
 
 from .config import settings
 
-if settings.redis_url == 'redis://fakeredis':
+if settings.redis_url.startswith('redis://fakeredis'):
```
patrickkwang, line 9: Is this change related to the others?

run.py
```diff
@@ -0,0 +1,7 @@
+import uvicorn
```
patrickkwang, line 1: I usually use a `main.sh` file to do this:
```bash
#!/usr/bin/env bash

export $(egrep -v '^#' .env | xargs)

# run api server
uvicorn strider.server:APP --host 0.0.0.0 --port 5781 --reload
```
I think I just prefer executables to be bash scripts than Python scripts, but are there other pros/cons to doing it either way?

alongreyber: The reason for the `run.py` script is so that it can be run with pdb. You can't do `python -m pdb main.sh` but you can do that with `run.py`. This is how FastAPI recommends debugging their applications: https://fastapi.tiangolo.com/tutorial/debugging/

strider/storage.py
alongreyber: Kind-of. When reading from a .env file Pydantic will add the port to a redis url, so "redis://fakeredis" becomes "redis://fakeredis:6379". That breaks this check here, so the change is necessary to make this PR work. Another alternative way to do this could be to add the port to this string.

patrickkwang: Okay, let's make this more explicit by adding the port or, if for some reason the port might change, doing e.g. `re.fullmatch(f"redis://fakeredis:\d+", settings.redis_url)`

patrickkwang (changes requested):

patrickkwang (approved):

PR #139, opened 2021-04-15
Implemented test client

alongreyber:
Switched from calling sync_query to using a `httpx.AsyncClient` for testing. The advantage is that we can catch cases here where our code produces an invalid response that should fail to validate. Previously, tests like this could pass incorrectly. 

Also removed the check for warnings in the output logs in favor of asserting that the response contains results.

### DIFF
--- a/tests/server/test_server.py
+++ b/tests/server/test_server.py
@@ -6,7 +6,6 @@ import json
 import os
 import httpx
 from fastapi.responses import Response
-from fastapi.testclient import TestClient
 
 from reasoner_pydantic import Query, Message, QueryGraph
 import pytest
@@ -28,8 +27,9 @@ settings.redis_url = "redis://fakeredis:6379/0"
 
 
 from strider.kp_registry import Registry
-from strider.server import APP, sync_query, generate_traversal_plan
+from strider.server import APP, generate_traversal_plan
 
+client = httpx.AsyncClient(app = APP, base_url="http://test")
 
 setup_logger()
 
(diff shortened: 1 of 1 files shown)


tests/server/test_server.py
```diff
@@ -376,11 +372,14 @@ async def test_log_level_param():
...
 
     # Check there are no debug logs
-    output = await sync_query(q, 'INFO')
+    response = await client.post("/query", json = q.dict())
+    output = response.json()
     assert not any(l['level'] == 'DEBUG' for l in output['logs'])
 
     # Check there are now debug logs
-    output = await sync_query(q, 'DEBUG')
+    response = await client.post("/query", json = q.dict(),
```
patrickkwang, line 380:
```suggestion
    response = await client.post("/query", json=q.dict(),
```

tests/server/test_server.py
```diff
@@ -376,11 +372,14 @@ async def test_log_level_param():
...
+                                 params = {"log_level" : "DEBUG"})
```
patrickkwang, line 381:
```suggestion
                                 params={"log_level": "DEBUG"})
```
Please use a PEP8 linter to catch these whitespace inconsistencies.

patrickkwang: This applies in many other places here as well.

patrickkwang (changes requested):
This is good stuff. Let's just make sure we're holding to style conventions.

patrickkwang (approved):

PR #140, opened 2021-04-15
Message merging

alongreyber:
This implements better message merging with custom merging functions for nodes and edges. Currently, edge merging is a bit simplistic (we don't merge edges like CHEBI-treats->MONDO and MONDO-treated_by->CHEBI) but still solves the core issue.

### DIFF
--- a/strider/trapi.py
+++ b/strider/trapi.py
@@ -6,7 +6,7 @@ import hashlib
 from reasoner_pydantic import Message, Result, QNode, Node, Edge, KnowledgeGraph
 from reasoner_pydantic.qgraph import QueryGraph
 
-from strider.util import deduplicate_by, WrappedBMT, get_from_all, \
+from strider.util import deduplicate_by, WrappedBMT, filter_none, get_from_all, \
     build_predicate_direction, extract_predicate_direction, \
     deduplicate, merge_listify, all_equal
 from strider.normalizer import Normalizer
@@ -49,7 +49,7 @@ def merge_nodes(knodes: list[Node]) -> Node:
     attributes_values = get_from_all(knodes, "attributes")
     if attributes_values:
         output_knode["attributes"] = \
-            merge_listify(attributes_values)
+            filter_none(merge_listify(attributes_values))
 
     return output_knode
 
@@ -61,7 +61,7 @@ def merge_edges(kedges: list[Edge]) -> Edge:
     attributes_values = get_from_all(kedges, "attributes")
     if attributes_values:
         output_kedge["attributes"] = \
-            merge_listify(attributes_values)
+            filter_none(merge_listify(attributes_values))
 
     predicate_values = get_from_all(kedges, "predicate")
     if not all_equal(predicate_values):
--- a/strider/trapi.py
+++ b/strider/trapi.py
@@ -2,11 +2,12 @@
 from collections import defaultdict
 import logging
 
-from reasoner_pydantic import Message, Result, QNode, Node, Edge
+from reasoner_pydantic import Message, Result, QNode, Node, Edge, KnowledgeGraph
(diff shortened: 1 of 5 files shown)


tests/server/test_server.py
```diff
@@ -1205,6 +1205,11 @@ async def test_solve_double_subclass():
         output["message"],
     )
 
+    # Check that the knowledge graph node has the correct category
+    assert set(output["message"]["knowledge_graph"]["nodes"]["MONDO:1"]["category"]) == \
```
patrickkwang, line 1209: At some point it might be nice to check this as part of `validate_message`.

strider/util.py
```diff
@@ -357,3 +357,43 @@ def add_cors_manually(APP, request, response, cors_options):
...
+def get_from_all(
+    dictionaries: list[dict],
+    key,
+    default=None
+):
+    """
+    Get list of values from dictionaries.
+    If it is not present in any dictionary, return the default value.
+    """
+    values = []
```
patrickkwang, line 370:
```suggestion
    values = [d[key] for d in dictionaries if key in d]
```

strider/trapi.py
```diff
@@ -283,3 +359,23 @@ def add_descendants(
...
+
+
+def build_unique_kg_edge_ids(message: Message):
+    """
+    Replace KG edge IDs with a string that represents
+    whether the edge can be merged with other edges
+    """
+    for edge_id in list(message["knowledge_graph"]["edges"].keys()):
+        edge = message["knowledge_graph"]["edges"].pop(edge_id)
+        new_edge_id = f"{edge['subject']}-{edge['predicate']}-{edge['object']}"
```
patrickkwang, line 371: In my experience, long edge ids like this make it annoying for this human to read a message. Would we lose anything by using a short hash instead? Aside from, theoretically, collisions.
```suggestion
        new_edge_id = hashlib.blake2b(f"{edge['subject']}-{edge['predicate']}-{edge['object']}".encode(), digest_size=6).hexdigest()
```
For 100,000 edges and a 6-byte hash, the chance of a collision is on the order of 1e-5, if my calculations are correct.

strider/trapi.py
```diff
@@ -283,3 +359,23 @@ def add_descendants(
...
 
     return qg
+
+
+def build_unique_kg_edge_ids(message: Message):
+    """
+    Replace KG edge IDs with a string that represents
+    whether the edge can be merged with other edges
+    """
+    for edge_id in list(message["knowledge_graph"]["edges"].keys()):
```
patrickkwang, line 369:
```suggestion
    # make a copy of the edge keys because we're about to change them
    for edge_id in list(message["knowledge_graph"]["edges"].keys()):
```

strider/trapi.py
```diff
@@ -28,26 +29,101 @@ def result_hash(result):
...
-    knodes = dict()
-    kedges = dict()
+
+    # Build knowledge graph edge IDs so that we can merge duplicates
+    for m in messages:
+        build_unique_kg_edge_ids(m)
+
+    kgraphs = [m["knowledge_graph"] for m in messages]
+
     results = []
```
patrickkwang, line 119:
```suggestion
    results = [r for m in messages for r in m["results"]]
```

patrickkwang (changes requested):

tests/server/test_server.py
alongreyber: I'm still trying to figure out how that would fit in to the format of the input to validate_message. If you have any ideas let me know.

patrickkwang: What if we used the same format for the knowledge_graph here as we do for the translator_overlay's kp_data?

patrickkwang (approved):

PR #147, opened 2021-04-21
Pinned-to-pinned query fixes

### LINKED ISSUE #133: 2-pinned-node query
```json
{
  "message": {
    "query_graph": {
      "edges": {
        "e01": {
            "subject": "n0",
            "object": "n1"
        }
      },
      "nodes": {
        "n0": {
          "category": "biolink:Disease",
          "id": "MONDO:0005301"
        },
        "n1": {
          "category": "biolink:Disease",
          "id": "MONDO:0017611"
        }
      }
    }
  }
}
```
Let's make sure that the results for this satisfy the query graph.

alongreyber:
Implemented pinned to pinned queries using results filtering. Fixes #133.

### DIFF
--- a/strider/trapi.py
+++ b/strider/trapi.py
@@ -5,10 +5,11 @@ import hashlib
 
 from reasoner_pydantic import Message, Result, QNode, Node, Edge, KnowledgeGraph
 from reasoner_pydantic.qgraph import QueryGraph
+from reasoner_pydantic.results import NodeBinding
 
-from strider.util import deduplicate_by, WrappedBMT, filter_none, get_from_all, \
+from strider.util import deduplicate_by, WrappedBMT, ensure_list, filter_none, get_from_all, \
     build_predicate_direction, extract_predicate_direction, \
-    deduplicate, merge_listify, all_equal
+    deduplicate, listify_value, merge_listify, all_equal
 from strider.normalizer import Normalizer
 from strider.config import settings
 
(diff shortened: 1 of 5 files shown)


strider/trapi.py
```diff
@@ -390,3 +391,112 @@ def build_unique_kg_edge_ids(message: Message):
...
+        if nb["id"] not in qgraph_node["id"]:
+            return False
+
+    kgraph_node = message["knowledge_graph"]["nodes"][nb["id"]]
+    if qgraph_node.get("category", None) is not None:
+        # Build list of allowable categories for kgraph nodes
+        qgraph_allowable_categories = []
+        for c in ensure_list(qgraph_node["category"]):
+            qgraph_allowable_categories.extend(
+                WBMT.get_descendants(c)
```
patrickkwang, line 411: Don't we have the "augmented" query graph where this is already done? Does it make sense to validate against that instead?

strider/trapi.py
```diff
@@ -390,3 +391,112 @@ def build_unique_kg_edge_ids(message: Message):
...
+            for qg_id, nb_list in result["node_bindings"].items()
+            for nb in nb_list
+        ) and all(
+            is_valid_edge_binding(message, eb, qgraph["edges"][qg_id])
+            for qg_id, eb_list in result["edge_bindings"].items()
+            for eb in eb_list
+        )
+    ]
+
+    remove_unbound_from_kg(message)
```
patrickkwang, line 476: This is slick.

strider/trapi.py
```diff
@@ -390,3 +391,112 @@ def build_unique_kg_edge_ids(message: Message):
...
+    """
+    if qgraph_node.get("id", None) is not None:
+        if nb["id"] not in qgraph_node["id"]:
+            return False
+
+    kgraph_node = message["knowledge_graph"]["nodes"][nb["id"]]
+    if qgraph_node.get("category", None) is not None:
+        # Build list of allowable categories for kgraph nodes
+        qgraph_allowable_categories = []
+        for c in ensure_list(qgraph_node["category"]):
```
patrickkwang, line 409: I think we already did `standardize_graph_lists` and shouldn't need to `ensure_list` here, right?

strider/trapi.py
```diff
@@ -390,3 +391,112 @@ def build_unique_kg_edge_ids(message: Message):
...
+    """
+    if qgraph_edge.get("id", None) is not None:
+        if eb["id"] not in qgraph_edge["id"]:
+            return False
+
+    kgraph_edge = message["knowledge_graph"]["edges"][eb["id"]]
+    if qgraph_edge.get("predicate", None) is not None:
+        # Build list of allowable predicates for kgraph edges
+        allowable_predicates = []
+        for p in ensure_list(qgraph_edge["predicate"]):
```
patrickkwang, line 436: If we add "predicate" to the `standardize_graph_lists` arguments, we wouldn't need to `ensure_list` here, right?

alongreyber: My understanding is that knowledge graph edge predicates are not allowed to be lists. We could technically do that but it seems like we should keep our dictionaries TRAPI compliant.

strider/trapi.py
alongreyber: The augmentation is currently done on the operation graph inside the query planner. Also, we can't use the `add_descendants` method because it only works on query graphs where predicates are lists. I could probably turn the add_descendants method into two methods (add_category_descendants and add_predicate_descendants) but I think at that point it's probably more complex than the current setup. If you have any thoughts on how to lay this out better let me know.

strider/trapi.py
alongreyber: The `standardize_graph_lists` is called on the knowledge graph, and `ensure_list` is working on a qgraph node.

patrickkwang (approved):

PR #148, opened 2021-04-21
Self edges fix

alongreyber:
Implemented self edge handing. Currently this isn't the prettiest solution so if you have any ideas on how to improve let me know.

Basic idea is to add in a second node for the KP request body with the suffix `.self` and then remove that suffix from bindings that we get back from the KP.

### DIFF
--- a/strider/fetcher.py
+++ b/strider/fetcher.py
@@ -336,8 +336,6 @@ def get_kp_request_body(
         request_edge["subject"] = source
         request_edge["object"] = target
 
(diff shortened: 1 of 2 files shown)


strider/fetcher.py
```diff
@@ -233,6 +234,15 @@ async def execute_step(
         ))
 
         for response in responses:
+
```
patrickkwang, line 237:
```suggestion
```

patrickkwang (approved):

PR #161, opened 2021-04-23
Identical attribute merging

### LINKED ISSUE #143: Merge identical attributes
Right now we appear to combine merged-node attribute lists by simply concatenating them. Some of these may be duplicates, or effectively duplicates. For now, we should at least handle attributes that are exact duplicates, and think about if/how we want to merge near-duplicates as well.

alongreyber:
Fixes #143.

### DIFF
--- a/strider/trapi.py
+++ b/strider/trapi.py
@@ -50,7 +50,9 @@ def merge_nodes(knodes: list[Node]) -> Node:
     attributes_values = get_from_all(knodes, "attributes")
     if attributes_values:
         output_knode["attributes"] = \
-            filter_none(merge_listify(attributes_values))
+            deduplicate_by(
+                filter_none(merge_listify(attributes_values)),
+                lambda d: tuple(d.items()))
 
     return output_knode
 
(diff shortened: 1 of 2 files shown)


strider/trapi.py
```diff
@@ -50,7 +50,9 @@ def merge_nodes(knodes: list[Node]) -> Node:
     attributes_values = get_from_all(knodes, "attributes")
     if attributes_values:
         output_knode["attributes"] = \
-            filter_none(merge_listify(attributes_values))
+            deduplicate_by(
+                filter_none(merge_listify(attributes_values)),
+                lambda d: tuple(d.items()))
```
patrickkwang, line 55: Technically, attribute `value`s are allowed to have arbitrary JSON types, including lists and dicts. I suspect that this will break in those cases. If we just need a hashable key, maybe `json.dumps(d, sort_keys=True)`?

patrickkwang (changes requested):

patrickkwang (approved):

PR #168, opened 2021-04-30
Convert Redis client to FastAPI dependency

patrickkwang:
This has a few benefits:
* Redis connections are tied to requests
* We don't need the configuration hack during testing
* We may be able to remove the Redis URL configuration entirely

and one drawback:
* We have to pass the client around everywhere

### DIFF
--- a/strider/storage.py
+++ b/strider/storage.py
@@ -6,12 +6,11 @@ import logging
 from typing import Optional
 
 import redis
-from redis import Redis
 
 from .config import settings
 
 
-def get_client() -> Redis:
+def get_client() -> redis.Redis:
     """Create a Redis client."""
     return redis.from_url(
         settings.redis_url,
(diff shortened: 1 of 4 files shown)


strider/storage.py
```diff
@@ -3,18 +3,17 @@
 import os
 import json
 import logging
+from typing import Optional
+
+import redis
+from redis import Redis
```
alongreyber, line 9: I don't think we need the double import, maybe just use `redis.Redis` below.

tests/test_server.py
```diff
@@ -14,14 +15,17 @@
...
+from strider.storage import get_client
 
 # Switch prefix path before importing server
 settings.kpregistry_url = "http://registry"
 settings.normalizer_url = "http://normalizer"
-settings.redis_url = "redis://fakeredis:6379/0"
-
-from strider.server import APP
 
+APP.dependency_overrides[get_client] = lambda: fakeredis.FakeRedis(
```
alongreyber, line 25: This is cool

alongreyber (changes requested):
This looks good. I don't love how verbose the FastAPI dependencies are, but I think it is still the right way to do this.

alongreyber (approved):

PR #178, opened 2021-05-11
Logging improvements

alongreyber:
- Added logging for all HTTP calls that Strider makes
- Added better logging output using extra properties on the LogEntry object
- Bumped reasoner pydantic version

### DIFF
--- a/strider/util.py
+++ b/strider/util.py
@@ -148,6 +148,9 @@ def log_response(r):
         "data" : r.text,
     }
 
+class StriderRequestError(BaseException):
+    """ Custom error indicating an issue with an HTTP request """
+
 async def post_json(url, request, logger, log_name):
     """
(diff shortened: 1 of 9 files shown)


strider/util.py
```diff
@@ -130,15 +131,59 @@ def entity_prefixes(self, entity):
...
+        })
+    except JSONDecodeError as e:
+        # Log error with response
+        logger.error({
+            "message": f"Received bad JSON data from {log_name}",
+            "request": e.request,
+            "response": e.response.text,
+            "error": str(e),
+        })
+    return None
```
patrickkwang, line 186: We seem to be treating this `None` output as reflecting an error state. Perhaps we should raise a custom error here instead.

patrickkwang (approved):

PR #187, opened 2021-05-21
Add workflow input handling and TRAPI 1.1 documentation

patrickkwang:


tests/test_server.py
```diff
@@ -1434,3 +1434,44 @@ async def test_registry_normalizer_unavailable():
...
+        "workflow": [
+            {
+                "id": "lookup",
+            },
+        ],
+    }
+
+    # Run
+    response = await client.post("/query", json=q)
+    output = response.json()
```
alongreyber, line 1477: This doesn't actually test anything because we have a middleware handler that catches exceptions and returns them as valid TRAPI. One check that should work is something like:
```python
assert output["message"]["results"]
```

alongreyber (changes requested):

alongreyber (approved):

PR #191, opened 2021-05-26
Add new logo to Swagger UI

patrickkwang:


strider/server.py
```diff
@@ -38,9 +38,16 @@
 
 LOGGER = logging.getLogger(__name__)
 
+DESCRIPTION = """
+<img src="/static/favicon.svg" width="300px">
```
alongreyber, line 42:
```suggestion
<img src="/static/favicon.svg" width="200px">
```

This seems a little large - I would rather not have to scroll down to see all of the routes.

alongreyber (changes requested):

PR #198, opened 2021-06-01
Added ability to handle structured logs from KPs

alongreyber:
This PR will check responses from KPs to see if they are LogEntries. If they are, the response is saved directly to the logs. This is useful for the integration with trapi-throttle.


alongreyber:
This needs to be redone due to the addition of trapi-throttle. It's low priority though, so I'm closing for now.

PR #207, opened 2021-06-10
v2.0.4

patrickkwang:
* Fix favicon path
* Exclude unset fields from KP queries
* Fix default KP registry URL
* Handle KP responses with no qgraph
* Update nodenorm URL to TRAPI 1.1 instance

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -614,11 +614,13 @@ async def test_kp_not_trapi():
     "http://ctd/query",
     Response(
         status_code=200,
-        content=json.dumps({"message": {
-            "query_graph": None,
-            "knowledge_graph": None,
-            "results": None,
-        }}),
+        content=json.dumps({
+            "message": {
+                "query_graph": None,
+                "knowledge_graph": None,
+                "results": None,
+            }
+        }),
     )
 )
 async def test_kp_response_empty():
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -649,7 +649,7 @@ async def test_kp_response_no_qg():
     response = await client.post("/query", json=q)
     output = response.json()
 
-    # Check that we stored the error
+    # Check that there are no logged errors.
     assert not output["logs"]
 
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -635,7 +635,11 @@ async def test_kp_response_empty():
     )
 
     # Create query
-    q = {"message" : {"query_graph" : QGRAPH}}
+    q = {
+        "message" : {
+            "query_graph" : QGRAPH
+        }
+    }
 
     # Run
     response = await client.post("/query", json=q)
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -622,9 +622,12 @@ async def test_kp_not_trapi():
         }),
     )
 )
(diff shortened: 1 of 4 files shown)


tests/test_server.py
```diff
@@ -593,6 +593,58 @@ async def test_kp_not_trapi():
     assert 'Received non-TRAPI compliant response from KP' in output['logs'][0]['message']
 
 
+@pytest.mark.asyncio
+@with_norm_overlay(settings.normalizer_url)
+@with_registry_overlay(
+    settings.kpregistry_url, {
+        'ctd': {
+            'url':
+            'http://ctd/query',
```
maximusunc, line 602:
```suggestion
            'url': 'http://ctd/query',
```

tests/test_server.py
```diff
@@ -593,6 +593,58 @@ async def test_kp_not_trapi():
...
+    })
+@with_response_overlay(
+    "http://ctd/query",
+    Response(
+        status_code=200,
+        content=json.dumps({"message": {
+            "query_graph": None,
+            "knowledge_graph": None,
+            "results": None,
+        }}),
```
maximusunc, line 621:
```suggestion
        content=json.dumps({
            "message": {
                "query_graph": None,
                "knowledge_graph": None,
                "results": None,
            }
        }),
```

tests/test_server.py
```diff
@@ -593,6 +593,58 @@ async def test_kp_not_trapi():
...
+        """
+        n0(( category biolink:ChemicalSubstance ))
+        n0(( id CHEBI:6801 ))
+        n0-- biolink:treats -->n1
+        n1(( category biolink:Disease ))
+        """
+    )
+
+    # Create query
+    q = {"message" : {"query_graph" : QGRAPH}}
```
maximusunc, line 638:
```suggestion
    q = {
        "message" : {
            "query_graph" : QGRAPH
        }
    }
```

tests/test_server.py
```diff
@@ -593,6 +593,58 @@ async def test_kp_not_trapi():
...
+    Response(
+        status_code=200,
+        content=json.dumps({"message": {
+            "query_graph": None,
+            "knowledge_graph": None,
+            "results": None,
+        }}),
+    )
+)
+async def test_kp_response_empty():
```
maximusunc, line 624: The naming and documentation of this test don't match with what you're asserting. Are no `logs` indicative of an empty response with null query graph?

patrickkwang: It's tricky because Strider is pretty resilient w.r.t. runtime errors. Prior to this fix, an error was encountered in our code when the KP response was missing the query graph, then the error was caught and was logged. This fixes the error and thus prevents the logged message.

tests/test_server.py
```diff
@@ -593,6 +593,66 @@ async def test_kp_not_trapi():
...
+        "message" : {
+            "query_graph" : QGRAPH
+        }
+    }
+
+    # Run
+    response = await client.post("/query", json=q)
+    output = response.json()
+
+    # Check that we stored the error
```
maximusunc, line 652:
```suggestion
    # Check that there are no logged errors.
```

maximusunc (approved):

PR #231, opened 2021-07-21
Prov

### LINKED ISSUE #204: Reformat provenance in line with Biolink 2.0
Reformat provenance in line with Biolink 2.0, consuming provenance from KPs and passing it along. Strider will then produce multiple provenance properties:
1. The provenance that came from the KP
2. Provenance of which KP it came from

**Due:** July 7, 2021

cbizon:
Updates the edge provenance strider provides.

KPs should be provide provenance about themselves, so it's no longer necessary to add where we got the edge.  Instead, we want to add provenance that aragorn passed the edge along.

Closes #204 

Couple notes:
* I added the attribute as aragorn rather than strider.  If we want to be more specific, we need to create a strider infores
* To get the tests to run, I had to modify some of the git locations in requirements-test, which I'm not sure is quite correct
* There is one test failing (test_solve_double_subclass) but it seems that it was failing on master already.


patrickkwang (approved):

patrickkwang:
All the tests pass on GitHub Actions, so I suspect something is funky with your local environment.

PR #237, opened 2021-07-28
Reformat provenance

uhbrar:
Added testing for provenance and adjusted edge attributes to report provenance for strider while also preserving provenance from KP's. Test ensures that provenance is included within edges.

Note:
Test may need to be adjusted if simple-kp/binder-kp appropriately reports KP name. Currently, it doesn't look like it does that correctly.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1660,6 +1660,7 @@ async def test_provenance():
     ]
     assert "infores:aragorn" in values
     assert "infores:kp0" in values
+    assert "infores:kp1" not in values
     assert "biolink:aggregator_knowledge_source" in attribute_type_ids
     assert "biolink:knowledge_source" in attribute_type_ids
     assert values.index("infores:aragorn") == attribute_type_ids.index("biolink:aggregator_knowledge_source")
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1609,3 +1609,47 @@ async def test_multiple_identifiers():
         },
         output["message"]
     )
+
+@pytest.mark.asyncio
+@with_translator_overlay(
+    settings.kpregistry_url,
+    settings.normalizer_url,
+    kp_data={
+        "kp0":
+        """
+            MONDO:0005148(( category biolink:Disease ))
+            MONDO:0005148<-- predicate biolink:treats --CHEBI:6801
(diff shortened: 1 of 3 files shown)


tests/test_server.py
```diff
@@ -1609,3 +1609,56 @@ async def test_multiple_identifiers():
...
+        n0-- biolink:related_to -->n1
+        """
+    )
+    q = {"message" : {"query_graph" : QGRAPH}}
+
+    # Run
+    response = await client.post("/query", json=q)
+    output = response.json()
+    edge = list(output["message"]["knowledge_graph"]["edges"].values())[0]
+    attributes = [
```
patrickkwang, line 1654: `attributes = edge["attributes"]`?

tests/test_server.py
```diff
@@ -1609,3 +1609,56 @@ async def test_multiple_identifiers():
...
+        attribute for attribute in edge["attributes"]
+    ]
+    provenance = [
+        [i["value"] for i in attributes],
+        [i["attribute_type_id"] for i in attributes]
+    ]
+    assert "infores:aragorn" in provenance[0]
+    assert "infores:None" in provenance[0]
+    assert "biolink:aggregator_knowledge_source" in provenance[1]
```
patrickkwang, line 1663: These are paired - "infores:aragorn" should be the "biolink:aggregator_knowledge_source" and "infores:kp0" should be the "biolink:knowledge_source".

tests/test_server.py
```diff
@@ -1609,3 +1609,56 @@ async def test_multiple_identifiers():
...
+
+    # Run
+    response = await client.post("/query", json=q)
+    output = response.json()
+    edge = list(output["message"]["knowledge_graph"]["edges"].values())[0]
+    attributes = [
+        attribute for attribute in edge["attributes"]
+    ]
+    provenance = [
+        [i["value"] for i in attributes],
```
patrickkwang, line 1658: It might be clearer to simply name these separately, e.g. "values" and "attribute_type_ids".

patrickkwang (changes requested):
* Let's get the fixed ~~simple-kp~~ binder in (#238), then rebase this
* Let's clean up the accidental venv history before this gets to master

uhbrar:
Made all changes, and cleaned up git history.

patrickkwang (approved):

PR #244, opened 2021-08-05
url for node norm has changed to nodenormalization-sri-dev.renci.org

PhillipsOwen:
like the title says


patrickkwang (approved):
Note that those configuration defaults can also be overridden by environment variables.

PR #245, opened 2021-08-05
v3.3

patrickkwang:


uhbrar (approved):
Looks good to me. It might be a good idea to bump the kp-registry version, but besides that, I don't see any problems. It also works a lot faster now, which is great.

PR #246, opened 2021-08-05
Biolink 2.1.0

patrickkwang:


cbizon (changes requested):
This all looks reasonable.  I am noticing that standup queries are still using ChemicalSubstance for input, so we might want to go the other direction as well. i.e. when a ChemicalSubstance comes in, we add SmallMolecule.

cbizon (approved):

PR #249, opened 2021-08-10
Use canonical predicates

patrickkwang:


strider/compatibility.py
```diff
@@ -87,16 +87,6 @@ async def map_prefixes(
...
-    async def map_curie(
-        self,
-        curie: str,
-        prefixes: dict[str, list[str]],
-    ) -> str:
-        """Map CURIE."""
-        await self.synonymizer.load_curies(curie)
-        curie_map = self.synonymizer.map(prefixes)
-        return curie_map.get(curie, curie)
-
```
uhbrar, line 99: Seems like this might be causing Github to be unable to build to run automatic checks.

requirements-test.txt
```diff
@@ -3,7 +3,7 @@ aiofiles==0.6.0
 redis==3.5.3
 
 # BMT + dependencies
-bmt-lite-2.1.0==1.0.2
+bmt-lite-2.1.0==2.1.1
```
uhbrar, line 6: This causing an error when trying to lock requirements for testing. 

`ERROR: Cannot install -r /app/requirements-test.txt (line 23) and bmt-lite-2.1.0==2.1.1 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies`

This is because binder 4.1.0 depends on bmt-lite<2.0.

uhbrar (changes requested):
Tests cannot be run locally due to these issues.

requirements-test.txt
patrickkwang: Thanks! I've upgraded binder.

uhbrar (approved):
Looks good now!

PR #253, opened 2021-08-18
Adding gunicorn to image

YaphetKG:


patrickkwang (changes requested):
So this will install gunicorn, but don't we need to edit the Dockerfile in order to actually use it?

YaphetKG:
> So this will install gunicorn, but don't we need to edit the Dockerfile in order to actually use it?

Hi Patrick, yes we can modify the entry command on prod to use Gunicorn , we are already doing something similar with the current uvicorn where we control number of workers and rootpath (1.1) . 

In the last push I added Uvicorn[standard] which installs some extra libs required by  gunicorn.

patrickkwang (approved):
Looks good. The only thing I'm not sure about is the dividing line between the ENTRYPOINT and CMD. If one were to provide a different command, I'm not sure that FastAPI works without the uvicorn worker (or a similar ASGI worker). Maybe the answer is just to move that option into the ENTRYPOINT.

davefol:
This is pretty straightforward and looks good to me @alongreyber. Think it's ready to merge.

alongreyber:
I'm testing locally and some stuff like `manage.py dev` isn't working. I'll request review when it's fixed.

alongreyber:
Had to fix some stuff that was preventing `manage.py dev` from working. Turns out gunicorn handles environment variables slightly differently from uvicorn.

PR #254, opened 2021-08-18
Handle qnode/qedge constraints

patrickkwang:


uhbrar (commented):
Some really weird behavior here, I think. Github is saying a test fails, but when I run it locally, there doesn't seem to be any issues. Not really sure what's going on with that, but it might be a good idea to look into it. 

Github is saying that the test that fails is in test_server.py: test_double_subclass. Not sure why it seems to be passing perfectly fine on my machine.

patrickkwang:
I'm concerned that this test is actually stochastic. For computational efficiency, we recently stopped merging knodes/kedges in the way we'd ideally like to. The category that comes out here depends on which KP responds first, which could hinge on vagaries of the asyncio event loop and aiosqlite. It's probably best just to change/remove that assertion.

uhbrar (approved):
With that change, this looks good to me now.

PR #266, opened 2021-09-02
Test without sleeping

patrickkwang:


uhbrar (approved):
Looks like this will solve the problem pretty well. I guess the wrapper is unnecessary for this.

PR #269, opened 2021-09-07
Random fixes

patrickkwang:


strider/compatibility.py
```diff
@@ -252,11 +252,11 @@ def map_curie(
             categories, identifiers = data[curie]
         except KeyError:
             return [curie]
-        prefixes = [
+        prefixes = {
             prefix
             for category in categories
             for prefix in prefixes.get(category, [])
-        ]
+        }
```
uhbrar, line 259: Not exactly sure why this should be a set instead of a list. Am I missing something here?

uhbrar (approved):
Overall, this looks good to me. Had a few questions, but it seems okay.

strider/compatibility.py
patrickkwang: It doesn't need to be ordered, there should be no duplicates, and we'd like to be able to search it quickly.

PR #280, opened 2021-09-11
Add profiler

### LINKED ISSUE #277: Add framework for debugging performance issues

alongreyber:
This PR adds a profiler that can be enabled with PROFILER=true environment variable. The profiler is available at localhost:5781/profiler and the generated profiles can be used with snakeviz for easy performance debugging. Fixes #277.

The only minor issue of this PR is that yappi is added to our deployed Dockerfile. There's not really a clean way to include it as a dev dependency (this would involve creating `requirements-dev.txt` and `requirements-dev-lock.txt` and `Dockerfile.dev`). I don't think it's going to be a big issue - yappi has no dependencies of its own and is not imported if PROFILER=true is not defined.

### DIFF
--- a/strider/profiler.py
+++ b/strider/profiler.py
@@ -1,0 +1,56 @@
+from pathlib import Path
+import time
+
+from fastapi.staticfiles import StaticFiles
+from fastapi.responses import HTMLResponse
+import yappi
+
+from .server import APP
+
+# Create directory to store request profile data
+PROFILE_DIRECTORY = "/profiles"
+Path(PROFILE_DIRECTORY).mkdir(exist_ok = True)
+
+# Middleware to capture profile data and save it
+# to files
+@APP.middleware("http")
+async def profiler_middleware(request, call_next):
+    # Run the request
+    with yappi.run():
+        call_result = await call_next(request)
+
+    # Save profile
+    stats = yappi.get_func_stats()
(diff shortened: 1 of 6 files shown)


strider/profiler.py
```diff
@@ -0,0 +1,65 @@
...
+    Wrapper on StaticFiles to serve files that are automatically
+    downloaded using the Content-Disposition header
+    """
+    def file_response(*args, **kwargs):
+        resp = StaticFiles.file_response(*args, **kwargs)
+        resp.headers["Content-Disposition"] = "attachment"
+        return resp
+
+# Create directory to store request profile data
+PROFILE_DIRECTORY = "/profiles"
```
patrickkwang, line 22: Do we mean to have an absolute path here?

patrickkwang (changes requested):
Can we add a section to the README about how to use this?

strider/profiler.py
alongreyber: Yes. This is in the docker container, so an absolute path in the root filesystem is fine.

patrickkwang: This _can_ be in a Docker container, but it can also be run locally if we choose a directory we have permission to create.

alongreyber: Who runs things outside of Docker?? Should be fixed now...

patrickkwang (approved):

PR #290, opened 2021-09-20
Remove ChemicalSubstance shims

### LINKED ISSUE #271: Remove ChemicalSubstance Alias
https://github.com/NCATSTranslator/minihackathons/issues/229

I think when we're sending SmallMolecule, we're also including ChemicalSubstance to try to smooth over the biolink 1-> 2 transition.

But now that everybody is on 2, at least one KP is complaining about the CS category, so I think we're ok to remove it?

patrickkwang:
Resolves #271.

### DIFF
--- a/strider/compatibility.py
+++ b/strider/compatibility.py
@@ -209,11 +209,7 @@ class Synonymizer():
 
         entities = [
             Entity(
-                entity["type"] + (
-                    ["biolink:ChemicalSubstance"]
-                    if "biolink:SmallMolecule" in entity["type"] else
-                    []
-                ),
+                entity["type"],
                 [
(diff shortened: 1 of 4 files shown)


maximusunc:
I think your changes look fine. However, when searching in the repo for `ChemicalSubstance`, I get 84 results. I see that your tests are all passing, but please confirm that all those search results are fine.

patrickkwang:
Yeah, ChemicalSubstance is commonly used in the tests, but most of the time we don't actually need to understand what it is. The tests I've fixed rely on traversing the biolink model somehow.

maximusunc (approved):

Issue #289, opened 2021-09-20
Handle node binding's `qnode_id` from Platers


patrickkwang:
`qnode_id` will be a field on node bindings (a sibling of `id`) that indicates which element in the qnode's `ids` is bound. We should use it to disambiguate unbatching.

PR #293, opened 2021-09-21
Unbatching

patrickkwang:


strider/trapi_throttle/trapi.py
```diff
@@ -66,7 +66,7 @@ def result_contains_node_bindings(
     """ Check that the result object has all bindings provided (qg_id->kg_id) """
     for qg_id, kg_ids in bindings.items():
         if not any(
-            nb["id"] in kg_ids
+            nb["id"] in kg_ids or nb.get("qnode_id") in kg_ids
```
maximusunc, line 69: Where is this `qnode_id` coming from? And is there a reason we combine the direct access and get method when looking for these values?

patrickkwang: This `qnode_id` business is an experimental feature for enabling unambiguous unbatching. It has been implemented by Automat. We use the direct access for `id` because it is a required field and we know it'll be there, but we use `get()` for `qnode_id` because most bindings will not have it - we'll get None instead of a KeyError.

patrickkwang:
Resolves #289.

maximusunc (approved):

maximusunc (approved):

PR #300, opened 2021-09-28
Add pinnedness planning

patrickkwang:


strider/query_planner.py
```diff
@@ -209,3 +210,78 @@ def get_query_graph_edge_kps(
...
+
+
+def get_next_qedge(qgraph):
+    """Get next qedge to solve."""
+    qgraph = copy.deepcopy(qgraph)
+    for qnode in qgraph["nodes"].values():
+        if qnode.get("ids") is not None:
+            qnode["ids"] = len(qnode["ids"])
+        else:
+            qnode["ids"] = 1_000_000
```
maximusunc, line 222: This could be `N` defined below

strider/query_planner.py
```diff
@@ -209,3 +210,78 @@ def get_query_graph_edge_kps(
...
+
+
+N = 1000000  # total number of nodes
+R = 25  # number of edges per node
+
+
+def get_pinnedness(qgraph, qnode_id):
+    """Get pinnedness of each node."""
+    adjacency_mat = get_adjacency_matrix(qgraph)
+    return -_get_pinnedness_recursively(
```
maximusunc, line 250: Why is there a minus sign here? I think it's easy to miss and should be more obvious.

strider/query_planner.py
```diff
@@ -209,3 +210,78 @@ def get_query_graph_edge_kps(
...
+        ids = qnode.get("ids")
+        if ids is None:
+            num_ids = N
+        elif isinstance(ids, list):
+            num_ids = len(ids)
+        else:
+            num_ids = ids
+        A[qnode_id][qnode_id] = num_ids
+    for qedge in qgraph["edges"].values():
+        A[qedge["subject"]][qedge["object"]] = True
```
maximusunc, line 285: What about self-edges?

maximusunc (changes requested):

strider/query_planner.py
patrickkwang: I modified some naming/documentation to try and make this more readable.

strider/query_planner.py
patrickkwang: 👍

strider/query_planner.py
patrickkwang: Good catch... this will break somewhat if self edges are present, and also will not work as we perhaps hope if there are multi-edges. I think it can be made to work nicely, it will just take a little more thought.

maximusunc (approved):

PR #301, opened 2021-09-29
Gene protein conflation

### LINKED ISSUE #257: Implement Gene-Protein Conflation

uhbrar:
Implements and tests gene protein conflation. Closes #257

### DIFF
--- a/strider/config.py
+++ b/strider/config.py
@@ -10,7 +10,7 @@ class Settings(BaseSettings):
     kpregistry_url: AnyUrl = "https://kp-registry.renci.org"
     omnicorp_url: AnyUrl = "http://robokop.renci.org:3210"
     biolink_url: AnyUrl = "https://bl-lookup-sri.renci.org"
-    normalizer_url: AnyUrl = "https://nodenormalization-sri-prod.renci.org/1.1"
+    normalizer_url: AnyUrl = "https://nodenormalization-sri.renci.org/docs"
 
     redis_url: RedisDsn = "redis://localhost"
     store_results_for: timedelta = timedelta(days=7)
--- a/strider/config.py
+++ b/strider/config.py
@@ -10,7 +10,7 @@ class Settings(BaseSettings):
     kpregistry_url: AnyUrl = "https://kp-registry.renci.org"
     omnicorp_url: AnyUrl = "http://robokop.renci.org:3210"
     biolink_url: AnyUrl = "https://bl-lookup-sri.renci.org"
-    normalizer_url: AnyUrl = "https://nodenormalization-sri.renci.org/docs"
+    normalizer_url: AnyUrl = "https://nodenormalization-sri.renci.org/"
 
     redis_url: RedisDsn = "redis://localhost"
     store_results_for: timedelta = timedelta(days=7)
--- a/strider/config.py
+++ b/strider/config.py
@@ -10,7 +10,7 @@ class Settings(BaseSettings):
     kpregistry_url: AnyUrl = "https://kp-registry.renci.org"
     omnicorp_url: AnyUrl = "http://robokop.renci.org:3210"
     biolink_url: AnyUrl = "https://bl-lookup-sri.renci.org"
-    normalizer_url: AnyUrl = "https://nodenormalization-sri-dev.renci.org/1.1"
+    normalizer_url: AnyUrl = "https://nodenormalization-sri-prod.renci.org/1.1"
 
     redis_url: RedisDsn = "redis://localhost"
     store_results_for: timedelta = timedelta(days=7)
--- a/strider/trapi.py
+++ b/strider/trapi.py
@@ -449,21 +449,21 @@ async def fill_categories_predicates(
                 node['categories'].append('biolink:Protein')
             if 'biolink:Protein' in node['categories'] and 'biolink:Gene' not in node['categories']:
                 node['categories'].append('biolink:Gene')
-            continue
-        if not isinstance(node_id, list):
-            node_id = [node_id]
+        else:
+            if not isinstance(node_id, list):
+                node_id = [node_id]
 
-        # Get full list of categorys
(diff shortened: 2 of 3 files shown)


strider/trapi.py
```diff
@@ -445,21 +445,25 @@ async def fill_categories_predicates(
     for node in qg['nodes'].values():
         node_id = node.get('ids', None)
         if not node_id:
-            continue
-        if not isinstance(node_id, list):
-            node_id = [node_id]
+            if 'biolink:Gene' in node['categories'] and 'biolink:Protein' not in node['categories']:
```
patrickkwang, line 448: A good practice is to touch as little code as possible so that a) PRs are easy, and b) git blames are helpful. Here we can just add these lines before the `continue` and not touch the rest.

strider/config.py
```diff
@@ -10,7 +10,7 @@ class Settings(BaseSettings):
     kpregistry_url: AnyUrl = "https://kp-registry.renci.org"
     omnicorp_url: AnyUrl = "http://robokop.renci.org:3210"
     biolink_url: AnyUrl = "https://bl-lookup-sri.renci.org"
-    normalizer_url: AnyUrl = "https://nodenormalization-sri-dev.renci.org/1.1"
+    normalizer_url: AnyUrl = "https://nodenormalization-sri-prod.renci.org/1.1"
```
patrickkwang, line 13: Is this the right URL? I thought that this was the one: https://nodenormalization-sri.renci.org/docs

tests/test_query.py
```diff
@@ -377,3 +377,69 @@ async def test_trivial_unbatching(redis):
...
+                n0 MONDO:0008114
+                n1 MESH:C035133
+                n2 HP:0007430
+            edge_bindings:
+                n0n1 MONDO:0008114-MESH:C035133
+                n1n2 MESH:C035133-HP:0007430
+            """
+        ],
+    },
+        output["message"])
```
patrickkwang, line 445: Check the formatting here.

tests/test_query.py
```diff
@@ -377,3 +377,69 @@ async def test_trivial_unbatching(redis):
...
+            MONDO:0008114(( category biolink:Disease ))
+            MONDO:0008114-- predicate biolink:related_to -->MESH:C035133
+            MESH:C035133(( category biolink:Gene ))
+        """,
+        "kp1":
+        """
+            MESH:C035133(( category biolink:Gene ))
+            MESH:C035133-- predicate biolink:related_to -->HP:0007430
+            HP:0007430(( category biolink:Protein ))
+            MESH:C035133-- predicate biolink:related_to -->HP:0007431
```
patrickkwang, line 397: What is the purpose of this?

patrickkwang (changes requested):

strider/trapi.py
uhbrar: I guess that's true. I'd always been told that continue statements were sort of "bad practice", so since the continue statement was no longer necessary, now that there was something actually in the if-block, I though it would be good to remove it.

tests/test_query.py
uhbrar: Good catch. I actually forgot that I put that in there. That was an incomplete part of the test that I must have forgotten about. I'll get that fixed now.

tests/test_query.py
```diff
@@ -377,3 +377,84 @@ async def test_trivial_unbatching(redis):
...
+        """
+    )
+
+    q = {
+        "message" : {"query_graph" : QGRAPH},
+        "log_level" : "ERROR"
+    }
+
+    output = await lookup(q, redis)
+
```
patrickkwang, line 436: Let's also test that the kgraph nodes contain both of the expected categories.

tests/test_query.py
```diff
@@ -377,3 +377,84 @@ async def test_trivial_unbatching(redis):
...
+            HP:0007430(( category biolink:Protein ))
+            HP:0007430-- predicate biolink:related_to -->CHEBI:6801
+            CHEBI:6801(( category biolink:Disease ))
+            HGNC:6284(( category biolink:Gene ))
+            HP:0007430-- predicate biolink:related_to -->HGNC:6284
+        """
+    },
+    normalizer_data="""
+        MONDO:0008114 categories biolink:Disease
+        HP:0007430 categories biolink:Protein
```
patrickkwang, line 409: We'll want to add both categories here to mock the new node normalizer with conflation.

tests/test_query.py
```diff
@@ -377,3 +377,84 @@ async def test_trivial_unbatching(redis):
...
+        MESH:C035133 categories biolink:Gene
+        HGNC:6284 categories biolink:Gene
+        CHEBI:6801 categories biolink:Disease
+        """
+)
+async def test_gene_protein_conflation(redis):
+    """Test conflation of biolink:Gene and biolink:Protein categories"""
```
patrickkwang, line 416: Please add documentation explaining why we have all of these edges and what each one is testing.

strider/config.py
```diff
@@ -10,7 +10,7 @@ class Settings(BaseSettings):
     kpregistry_url: AnyUrl = "https://kp-registry.renci.org"
     omnicorp_url: AnyUrl = "http://robokop.renci.org:3210"
     biolink_url: AnyUrl = "https://bl-lookup-sri.renci.org"
-    normalizer_url: AnyUrl = "https://nodenormalization-sri-dev.renci.org/1.1"
+    normalizer_url: AnyUrl = "https://nodenormalization-sri.renci.org/docs"
```
patrickkwang, line 13: I think we just want to remove the /docs and this will be good. But let's test it with the deployed normalizer.

patrickkwang (changes requested):

tests/test_query.py
uhbrar: Since we aren't adding categories to the kgraph, this is no longer relevant.

patrickkwang (approved):

PR #312, opened 2021-10-22
Benchmarking

alongreyber:
Added command `manage.py benchmark` to run a set of benchmarks. Currently benchmarks the `merge_messages` function.


patrickkwang (dismissed):
Looks pretty good. Some minor formatting issues. Also, how important are each of these extra dependencies?

davefol (approved):

PR #331, opened 2021-11-10
increasing async write timeout

PhillipsOwen:
Strider is throwing httpx:WriteTimeout and/or httpx:ReadTimeout errors when posting results to a callback when using the /asyncquery endpoint. 

Some queries create results that may be in excess of 100mb. it may be in some cases that the data cannot be transferred successfully within the default timeout period (5 seconds). 

the method "async def server.py/async_lookup()" is used to respond to the callback with results that strider has accumulated. in this function the call to "httpx.AsyncClient()" does not have a timeout specified. the modification suggested in this PR increases the timeout to 10 minutes which, in my experience, works well in high latency networks.

please see the enclosed query graph as an example input. to the strider /asyncquery endpoint

[creates-large-resultset.json.txt](https://github.com/ranking-agent/strider/files/7514492/creates-large-resultset.json.txt)

### DIFF
--- a/strider/server.py
+++ b/strider/server.py
@@ -269,7 +269,7 @@ async def async_lookup(
 ):
     """Preform lookup and send results to callback url"""
     query_results = await lookup(query_dict, redis_client)
-    async with httpx.AsyncClient() as client:
+    async with httpx.AsyncClient(timeout=httpx.Timeout(timeout=600.0)) as client:
         await client.post(callback, json=query_results)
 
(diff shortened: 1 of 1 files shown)


cbizon:
Just a note that we had to make an analogous change in the aragorn code that posts async results back to the ARS

alongreyber (approved):

kennethmorton (approved):

PR #333, opened 2021-11-15
Add maturity list to kp registry search

maximusunc:
Add hard-coded list of allowed maturities to kp registry search call based on own maturity level. This is pretty hacky to get it working quickly.

Corresponding PR: https://github.com/ranking-agent/kp-registry/pull/26

@cbizon

### DIFF
--- a/strider/kp_registry.py
+++ b/strider/kp_registry.py
@@ -7,6 +7,7 @@ from typing import Union
 import httpx
 
 from strider.util import WBMT
+from strider.config import settings
 
 
 class Registry():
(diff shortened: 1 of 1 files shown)


strider/kp_registry.py
```diff
@@ -96,6 +97,10 @@ async def search(
             for desc in WBMT.get_descendants(inverse)
         ]
         object_categories = [desc for cat in object_categories for desc in WBMT.get_descendants(cat)]
+        if settings.openapi_server_maturity == "development":
```
alongreyber, line 100: It looks like based on the SmartAPI docs, maturity also includes the value "staging." Does that need to be included here?

alongreyber (approved):
I'm not super up-to-date on the x-maturity additions but this looks reasonable to me.

strider/kp_registry.py
maximusunc: No, this is a pretty quick patch. I'll open an issue where we can revisit how this is done and to add staging if need be.

PR #357, opened 2022-02-28
Fix independent results bug

adamviola:
Fixes a bug in `fetcher.py` that caused unnecessary nodes to appear in kgraphs provided by `lookup()`. The fix helps merge time.

The fetcher was unintentionally adding nodes and edges to the same intermediate kgraph used across multiple results, causing the size of the kgraph to grow with each subsequent result. We're now deepcopying the initial intermediate kgraph for separate use in each result.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -2100,6 +2100,7 @@ async def test_large_merging_performance(client):
     # Check that we only have one result
     assert len(output["message"]["results"]) == 1
 
+
 @pytest.mark.asyncio
 @with_translator_overlay(
     settings.kpregistry_url,
@@ -2118,7 +2119,7 @@ async def test_large_merging_performance(client):
         """
     },
 )
-async def test_yield_indepedent_results(client):
+async def test_yield_independent_results(client):
     """Test that when we ask for a solution, it's provided as independent results that are merged together"""
     QGRAPH = query_graph_from_string(
         """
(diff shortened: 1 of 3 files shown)


tests/test_server.py
```diff
@@ -2098,3 +2099,52 @@ async def test_large_merging_performance(client):
...
+            HP:0004324(( category biolink:PhenotypicFeature ))
+            HP:0004325(( category biolink:PhenotypicFeature ))
+            CHEBI:6801-- predicate biolink:treats -->MONDO:0005148
+            CHEBI:6801-- predicate biolink:treats -->MONDO:0005149
+            MONDO:0005148-- predicate biolink:has_phenotype -->HP:0004324
+            MONDO:0005149-- predicate biolink:has_phenotype -->HP:0004325
+        """
+    },
+)
+async def test_yield_indepedent_results(client):
```
uhbrar, line 2121: Can you fix the typo in this line?

uhbrar (changes requested):
Also, some tests are failing and some reformatting is necessary. We should make sure all tests pass before we merge these changes.

tests/test_server.py
```diff
@@ -2098,3 +2099,52 @@ async def test_large_merging_performance(client):
...
+    )
+    await binder.setup(QGRAPH)
+
+    async with binder:
+        output = [(kgraph, result) async for kgraph, result in binder.lookup(None)]
+
+    # 2 results
+    assert len(output) == 2
+
+    breakpoint()
```
uhbrar, line 2146: This breakpoint should be removed

uhbrar (changes requested):

kennethmorton (approved):

uhbrar (changes requested):
Once we bring this branch up to date, it should be ready to merge in.

uhbrar (approved):

Issue #354, opened 2022-02-10
Crash with certain query.


PhillipsOwen:
The enclosed query seems to crash Strider (enclosed), versio ghcr.io/ranking-agent/strider:v3.17.0. 

Here is the stack trace:

2022-02-10T09:54:16-05:00 Traceback (most recent call last):
2022-02-10T09:54:16-05:00   File "/app/./strider/trapi_throttle/throttle.py", line 44, in wrapper
2022-02-10T09:54:16-05:00     return await fcn(*args, **kwargs)
2022-02-10T09:54:16-05:00   File "/app/./strider/trapi_throttle/throttle.py", line 288, in process_batch
2022-02-10T09:54:16-05:00     filtered_msg = filter_by_curie_mapping(
2022-02-10T09:54:16-05:00   File "/app/./strider/trapi_throttle/trapi.py", line 98, in filter_by_curie_mapping
2022-02-10T09:54:16-05:00     filtered_msg.kgraph = KnowledgeGraph(
2022-02-10T09:54:16-05:00   File "pydantic/main.py", line 422, in pydantic.main.BaseModel.__setattr__
2022-02-10T09:54:16-05:00 ValueError: "Message" object has no field "kgraph"
2022-02-10T09:54:24-05:00 Traceback (most recent call last):
2022-02-10T09:54:24-05:00   File "/app/./strider/trapi_throttle/throttle.py", line 44, in wrapper
2022-02-10T09:54:24-05:00     return await fcn(*args, **kwargs)
2022-02-10T09:54:24-05:00   File "/app/./strider/trapi_throttle/throttle.py", line 288, in process_batch
2022-02-10T09:54:24-05:00     filtered_msg = filter_by_curie_mapping(
2022-02-10T09:54:24-05:00   File "/app/./strider/trapi_throttle/trapi.py", line 99, in filter_by_curie_mapping
2022-02-10T09:54:24-05:00     nodes={
2022-02-10T09:54:24-05:00   File "/app/./strider/trapi_throttle/trapi.py", line 100, in <dictcomp>
2022-02-10T09:54:24-05:00     binding["id"]: message.knowledge_graph.nodes[binding["id"]]
2022-02-10T09:54:24-05:00 TypeError: 'NodeBinding' object is not subscriptable

[chris-mondo-gene-small.json.txt](https://github.com/ranking-agent/strider/files/8042514/chris-mondo-gene-small.json.txt)

uhbrar:
[...]

PR #358, opened 2022-03-02
use correct pydantic model fields in trapi throttle

### LINKED ISSUE #353: No results for two hop?
This query goes (hyperlipidemia)-(gene)-(small molecule), with related_tos for the edges.

For some reason, (dev) strider returns no results.   I can't see why not - searching robokopkg there should be plenty of such paths.

```
qq={
  "message": {
    "query_graph": {
      "nodes": {
        "n0": {
          "ids": ["MONDO:0021187"]
        },
        "n1": {
          "categories": [
            "biolink:Gene"
          ]
        },
        "n2": {
          "categories": [
            "biolink:SmallMolecule"
          ]
        }
      },
      "edges": {
        "e01": {
          "object": "n0",
          "subject": "n1",
          "predicates": ["biolink:related_to"]
        },
        "e02": {
          "subject":"n1",
          "object": "n2",
          "predicates": ["biolink:related_to"]
        }
      }
    }
  }
}
```

Maybe something is wrong with the query, but I don't see what, and I got results back for a different input ID (hypothyroidism)

uhbrar:
This fixes an issue that is present among some multi-hop queries by properly utilizing pydantic models in trapi throttle. This should also close #353 and #354. 

However, this does bring up another pressing issue. When this error was present, strider would error out and then hang until reset. We need to have more robust error handling to ensure that this is not an issue in the future.

### DIFF
--- a/strider/trapi_throttle/trapi.py
+++ b/strider/trapi_throttle/trapi.py
@@ -95,15 +95,15 @@ def filter_by_curie_mapping(
     }
 
     # Construct result-specific knowledge graph
-    filtered_msg.kgraph = KnowledgeGraph(
+    filtered_msg.knowledge_graph = KnowledgeGraph(
         nodes={
-            binding["id"]: message.knowledge_graph.nodes[binding["id"]]
+            binding.id: message.knowledge_graph.nodes[binding.id]
             for result in filtered_msg.results
             for _, bindings in result.node_bindings.items()
             for binding in bindings
         },
         edges={
-            binding["id"]: message.knowledge_graph.edges[binding["id"]]
+            binding.id: message.knowledge_graph.edges[binding.id]
             for result in filtered_msg.results
             for _, bindings in result.edge_bindings.items()
             for binding in bindings


kennethmorton:
Was this issue caught through offline testing? If so can we add a unit test to catch this in the future?

It certainly looks like a bug, but why didn't we catch it?

uhbrar:
A unit test can't catch this because it uses the batching code that directly deals with how we call external services. If we want to test this, we would have to do one of two things:
1. Implement tests using offline services; or
2. Create a mocked test that actually results in batched requests

Option 1 is bad for reasons we have discussed previously and since maintaining tests that call external services is a massive headache. Option 2 sounds reasonable but would require someone to sit down and create a multi-hop test that uses many KP's that have extensive knowledge and would provide dozens of curies, which is where the batching actually kicks in. Unfortunately, both of these options seem pretty infeasible. 

That being said, we should have more robust error handling so we don't have situations where strider hangs because of bad code, bad requests, or some other type of error. This was fine because it had never been deployed except to dev, if this had been put into prod, it would've been a massive issue.

kennethmorton:
I can accept that not all things can be tested. You still must take the canary to the coal mine. And that's why we have dev. Let's cut an image and keep testing this one.

PR #360, opened 2022-03-07
EPC now incudes value_type_id and attribute_source

kennethmorton:
This simple change puts us in line with the the discussion here:

https://github.com/ranking-agent/aragorn/issues/86#issuecomment-1059470372


uhbrar (approved):
After reading through the documentation, this looks like the appropriate changes to make right now.

Issue #362, opened 2022-03-14
bad query causes strider 500


cbizon:
```
query={
    "message": {
        "query_graph": {
            "edges": {
                "e00": {
                    "subject": "n01",
                    "object": "porphyria"
                }
            },
            "nodes": {
                "porphyria": {
                    "ids": [
                        "MONDO:0037939"
                    ]
                },
                "n01": {
                    "ids": [
                        "biolink:NamedThing"
                    ]
                }
            }
        }
    }
}
```

There is an error in the above query: n01 has an ids element when it means to have a categories element. 

However, this isn't malformed, it's just an id that strider won't be able to resolve.

In this case, strider is throwing a 500.

PR #364, opened 2022-03-23
500 on bad query

uhbrar:
This PR fixes the strider 500 on a bad query as mentioned in #362.

However, doing this also has some odd effects. For one thing, we actually do get some results now, which doesn't make sense to me. I don't really know how or why any KP would be sending us anything, but we are receiving response with 23 results back from Genetics KP. This clearly shouldn't be happening.

### DIFF
--- a/strider/normalizer.py
+++ b/strider/normalizer.py
@@ -29,7 +29,7 @@ class Normalizer:
 
         types = []
         for c in curies:
-            if results[c] is None:
+            if results.get(c) is None:
                 self.logger.warning(f"Normalizer knows nothing about {c}")
                 continue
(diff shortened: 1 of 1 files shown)


cbizon:
For the genetics KP, can you make an issue on their repo please?

kennethmorton (approved):

uhbrar:
This actually looks like a repeat of this same issue: https://github.com/broadinstitute/genetics-kp-dev/issues/26

@cbizon Should I make a new issue anyway? I'm pretty sure resolving that earlier issue will address this as well.

kennethmorton:
@uhbrar Just tack on a message with a reference back here.

PR #371, opened 2022-05-19
Fix hanging bug

maximusunc:
There are a few things that this PR takes care of:
- Even when there were no curies in the message, we were sending an empty list to Node Norm, who would error. There are now checks for an empty list and we won't send the request.
- We've bumped reasoner-pydantic versions to fix some validation issues
- I removed the pinned kp-registry and binder versions in our testing requirements so we're using the most recent. I also updated the tests because they were using an old kp-registry version.
- To fix the strider hanging issue, I added a debug call stack so we can hopefully find any hangs in the future easier. I also added a generic `Exception` handler to the throttle so we handle errors more gracefully.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,9 +1,9 @@
 asgiar==0.2.0
 fakeredis==1.5.0
-kp-registry @ git+https://github.com/ranking-agent/kp_registry@v2.2.0
+kp-registry @ git+https://github.com/ranking-agent/kp_registry
 pytest==6.2.4
 pytest-asyncio==0.15.1
 pytest-cov==2.12.1
-binder @ git+https://github.com/TranslatorSRI/binder@v4.3.2
+binder @ git+https://github.com/TranslatorSRI/binder
 orjson==3.6.0
 tqdm==4.62.3
--- a/strider/trapi_throttle/throttle.py
+++ b/strider/trapi_throttle/throttle.py
@@ -288,9 +288,7 @@ class ThrottledServer:
                             filtered_msg = filter_by_curie_mapping(
                                 message, curie_mapping, kp_id=self.id
                             )
-                            filtered_msg.query_graph = QueryGraph.parse_obj(
-                                request_value_mapping[request_id]
-                            )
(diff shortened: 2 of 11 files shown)


requirements-test.txt
```diff
@@ -1,9 +1,9 @@
 asgiar==0.2.0
 fakeredis==1.5.0
-kp-registry @ git+https://github.com/ranking-agent/kp_registry@v2.2.0
+kp-registry @ git+https://github.com/ranking-agent/kp_registry
 pytest==6.2.4
 pytest-asyncio==0.15.1
 pytest-cov==2.12.1
-binder @ git+https://github.com/TranslatorSRI/binder@v4.3.2
+binder @ git+https://github.com/TranslatorSRI/binder
```
kennethmorton, line 7: Should we be locking these down?

strider/trapi_throttle/throttle.py
```diff
@@ -266,43 +266,53 @@ async def process_batch(
...
-                            )
+                            filtered_msg.query_graph = request_value_mapping[
+                                request_id
+                            ].message.query_graph.copy()
                             response_values[request_id] = ReasonerResponse(
                                 message=filtered_msg
                             )
-                        except BatchingError as err:
-                            # the response is probably malformed
-                            response_values[request_id] = err
```
kennethmorton, line 299: Just leaving a note that this line is the critical line. As this was previously written it was possible to error out with a different exception. If you don't call this line you get stuck in an infinite while. In the edits here the try is restructured so that we will always fall to the except below for every exception.

kennethmorton (approved):

cbizon (approved):

requirements-test.txt
maximusunc: I had a little internal debate here as well. Where I landed was that these were kind of redundant with the lock files sitting right next to these. I figured not pinning them here would help keep Strider using the latest versions of these dependencies (as was needed with the kp-registry, which was two minor versions behind).

kennethmorton: Cool. You thought about it and I like that.

PR #379, opened 2022-06-29
add error handling for bouncing callback

uhbrar:


strider/server.py
```diff
@@ -357,10 +357,14 @@ async def multi_lookup(callback, queries: dict, query_keys: list, redis_client:
...
         query_result = await lookup(queries[query_key], redis_client)
-        async with httpx.AsyncClient(timeout=httpx.Timeout(timeout=600.0)) as client:
-            await client.post(callback, json=query_result)
-
-    await asyncio.gather(*map(single_lookup, query_keys))
+        try:
+            async with httpx.AsyncClient(timeout=httpx.Timeout(timeout=600.0)) as client:
+                await client.post(callback, json=query_result)
+        except Exception as e:
+            LOGGER.debug(e)
```
maximusunc, line 366: I think this should be .error()

strider/server.py
```diff
@@ -357,10 +357,16 @@ async def multi_lookup(callback, queries: dict, query_keys: list, redis_client:
...
-
-    await asyncio.gather(*map(single_lookup, query_keys))
+        try:
+            async with httpx.AsyncClient(
+                timeout=httpx.Timeout(timeout=600.0)
+            ) as client:
+                await client.post(callback, json=query_result)
+        except Exception as e:
+            LOGGER.debug(e)
+            raise e
```
maximusunc, line 367: Do we want to raise an error here? Or just continue? How do we want to handle failed callbacks?

maximusunc (changes requested):

strider/server.py
uhbrar: Considering how we'd talked about using the `return_exceptions=True` flag in the gather on line 369, I thought raising it made the most sense. But maybe we just want to throw it in the logs and continue. I'm open to either.

strider/server.py
uhbrar: That's true

maximusunc (approved):

PR #383, opened 2022-08-10
setup logging

cbizon:


maximusunc:
Not sure why this wasn't already set up. What are we hoping to gain out of this? Saving logs to a file? Will this work on ITRB?

cbizon:
Not sure why it wasn't set up already.  I was really trying to get the info logs to show up on the stdout log.   Turning on this at least seems to work locally

uhbrar:
We've enabled it in the past for debugging specific issues, but we've never included it in an actual release, at least not one that I remember.

PR #385, opened 2022-09-06
Trapi1.3

uhbrar:
Merges TRAPI 1.3 code


Dockerfile
```diff
@@ -1,5 +1,9 @@
+# Use RENCI python base image
 FROM renciorg/renci-python-image:v0.0.1
 
+# Build from this branch. Default to master
+ARG BRANCH_NAME=master
```
maximusunc, line 5: Can this be removed?

strider/server.py
```diff
@@ -50,7 +50,7 @@
     title="Strider",
     description=DESCRIPTION,
     docs_url=None,
-    version="3.18.4",
+    version="4.0.1",
```
maximusunc, line 53: Do we need to bump the version?

uhbrar: Good catch. I was thinking we don't need to bump the version, but I forgot that we had made changes to main since the last release.

Dockerfile
uhbrar: I think it can be, yes.

requirements.txt
```diff
@@ -13,7 +13,7 @@ uvicorn[standard]==0.14.0
 jsonpickle==1.4.2
 python-dotenv==0.17.0
 fastapi==0.65.2
-reasoner-pydantic==2.2.2
+reasoner-pydantic==3.0.0
```
kennethmorton, line 16: reasoner-pydantic==3.0.1 ?

uhbrar: updated

kennethmorton (approved):

PR #386, opened 2022-09-06
Max process time

uhbrar:
Adds max_process_time

### DIFF
--- a/strider/server.py
+++ b/strider/server.py
@@ -356,7 +356,7 @@ async def async_lookup(
             lookup(query_dict, redis_client), timeout=max_process_time
         )
     except asyncio.TimeoutError:
-        LOGGER.info("Process cancelled due to timeout.")
+        LOGGER.warning("Process cancelled due to timeout.")
         query_results = {
             "message": {},
             "status_communication": {"strider_process_status": "timeout"},
@@ -378,7 +378,7 @@ async def multi_lookup(callback, queries: dict, query_keys: list, redis_client: 
                 lookup(queries[query_key], redis_client), timeout=max_process_time
             )
         except asyncio.TimeoutError:
-            LOGGER.info("Process cancelled due to timeout.")
+            LOGGER.warning("Process cancelled due to timeout.")
         try:
             async with httpx.AsyncClient(
                 timeout=httpx.Timeout(timeout=600.0)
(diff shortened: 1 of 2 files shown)


strider/server.py
```diff
@@ -348,7 +350,12 @@ async def async_lookup(
     redis_client: Redis,
 ):
     """Perform lookup and send results to callback url"""
-    query_results = await lookup(query_dict, redis_client)
+    try:
+        query_results = await asyncio.wait_for(
+            lookup(query_dict, redis_client), timeout=max_process_time
+        )
+    except asyncio.TimeoutError:
```
maximusunc, line 357: Would this fail on line 353 because `query_results` is undefined? We should send some sort of timeout message back to aragorn. Same with the other functions

uhbrar: I'm not sure what the timeout message should be. We do have usage of the "strider_commuincation" field, so we could add something using that.

uhbrar:
I've added a process status communication to the message being sent. Since multiquery tells ARAGORN when all processes are finished, and one of them is killed in this process, multiquery already sends a "finished" signal. Therefore, it's not necessary for multiquery to also utilize this timeout signal.

strider/server.py
```diff
@@ -348,16 +350,35 @@ async def async_lookup(
...
-    query_results = await lookup(query_dict, redis_client)
-    async with httpx.AsyncClient(timeout=httpx.Timeout(timeout=600.0)) as client:
-        await client.post(callback, json=query_results)
+    query_results = {}
+    try:
+        query_results = await asyncio.wait_for(
+            lookup(query_dict, redis_client), timeout=max_process_time
+        )
+    except asyncio.TimeoutError:
+        LOGGER.info("Process cancelled due to timeout.")
```
maximusunc, line 359: I would change these to `warning`

strider/server.py
```diff
@@ -348,16 +350,35 @@ async def async_lookup(
...
 
     async def single_lookup(query_key):
-        query_result = await lookup(queries[query_key], redis_client)
+        query_result = {}
+        try:
+            query_result = await asyncio.wait_for(
+                lookup(queries[query_key], redis_client), timeout=max_process_time
+            )
+        except asyncio.TimeoutError:
+            LOGGER.info("Process cancelled due to timeout.")
```
maximusunc, line 381: Even if Strider sends Aragorn an "all done", I think it would still be useful to see on Aragorn's side a response for all queries and any timeout messages.

maximusunc (changes requested):

maximusunc (approved):

PR #400, opened 2023-04-06
Fix tests

maximusunc:
This fixes all the tests, and also fixes a couple (crucial) bugs in current Strider.
- Our canonical edge logic was messed up since bumping our bmtlite version
- Our kp lookup for inverse predicates was also messed up

This PR has a lot of commits because we haven't merged to master since adding kp registry to redis, which broke all the tests. Hopefully this puts us back on track and we should keep tests up to date going forward.


cbizon (approved):
There's a lot here, and I'm no strider expert, but I don't see anything alarming.

kennethmorton (approved):
Didn't run, but nothing looks alarming. Test clean up was long over due, and this branch has good milage on it. Approved.

PR #404, opened 2023-05-25
Fix analyses

uhbrar:
Fixes issue where multiple analyses were being generated for each result, as referenced in [this issue](https://github.com/ranking-agent/aragorn/issues/177).


strider/server.py
```diff
@@ -465,6 +465,9 @@ async def lookup(
             # Update the results
             output_results.update(result_message.results)
 
+    for result in output_results:
+        result.combine_analyses_by_resource_id()
```
kennethmorton, line 469: This function is gonna get called a lot when there are lots of results. Can we reduce the costs here by short cutting if this call is even required?

strider/compatibility.py
```diff
@@ -180,6 +180,10 @@ def add_source(message: Message, kp_id):
                 "upstream_resource_ids": [kp_id],
             }
         )
+    for result in message.get("results", []):
+        for analysis in result.get("analyses", []):
+            analysis["resource_id"] = "infores:aragorn"
+            analysis["score"] = None
```
maximusunc, line 186: I would put a note here maybe saying that we end up clearing out any scores given to us by KPs.

strider/server.py
uhbrar: Max and I talked about this, and we couldn't think of a good way to reduce the load here since the results need to be checked for matching analyses after all of the results are collected. We have to support a use case where the analyses are kept separate, otherwise we could just do it as part of the update function in the line above.

cbizon: I guess you could check to see if there's only one analysis and save a function call sometimes, but I'd want to profile it to see if it matters.

uhbrar: We could do this. I suspect it wouldn't really have much of an effect since the bottleneck is looping through all of the results, but it certainly wouldn't hurt.

cbizon: Maybe the analyses can instead be combined on addition to the same result or when results are merged?

maximusunc: Just to throw some metrics in here, I've run a query with 8901 results through without any combining, with combining on all results, and then with the analysis length check, and they all finished in the same amount of time, so right now I'm not too worried about this.

uhbrar: Thanks Max. I think that means we'll probably be safe merging this in for now then. We can revisit other strategies if we think it's a bottleneck in the future.

maximusunc (approved):

PR #416, opened 2023-07-07
Information Content filtering

maximusunc:


strider/fetcher.py
```diff
@@ -128,6 +128,7 @@ async def generate_from_kp(
             )
             onehop_response = await kp.solve_onehop(
                 onehop_qgraph,
+                last_hop=len(qgraph["edges"]) == 1,
```
kennethmorton, line 131: I understand why we want to keep track of last_hop, but I dont understand this logic in the check. There is only one qedge?

strider/trapi.py
```diff
@@ -166,6 +167,51 @@ def is_ancestor(a, b):
...
+    """Filter all nodes based on information content."""
+    keep_edges = []
+    new_results = Results.parse_obj([])
+    for result in message.results or []:
+        keep = True
+        for node_bindings in result.node_bindings.values():
+            for node_binding in node_bindings:
+                curie = curie_map.get(node_binding.id)
+                if (
+                    curie is not None
```
kennethmorton, line 185: Is there a legitimate reason this could happen? I get it as a safety, but do we need it? How about the opposite as a safety?

kennethmorton (approved):

strider/fetcher.py
maximusunc: As we work through the query one hop at a time, we're popping off each qedge as we go. This checks that we're on to the last hop.

strider/trapi.py
maximusunc: I think there's a good probability that this can happen occasionally. The `curie_map` is all of the nodes that the node normalizer knows about in the given message, but there's a chance that a node in some kp's response isn't known by node norm.

PR #421, opened 2023-07-17
Umls filtering

maximusunc:


strider/trapi.py
```diff
@@ -193,6 +193,10 @@ def filter_information_content(
                 if (
                     curie is not None
                     and curie.information_content < information_content_threshold
+                ) or (
```
cbizon, line 196: So my assumption here is that `curie is None` means that nodenorm doesn't know about the UMLS.   I think that a better condition (even more filtering) is that OR nodenorm does know about it and it only has a UMLS ID.

maximusunc: Added.

cbizon (approved):

PR #427, opened 2023-08-07
Timeout param

maximusunc:
Adds a check for `timeout_seconds` inside of request `parameters` sent from Aragorn. This will enable Strider to take longer for kp lookups to provide more thorough answers that get cached in Aragorn.

We're also adding a `/clear_cache` endpoint so someone can easily clear the onehop cache.

### DIFF
--- a/strider/knowledge_provider.py
+++ b/strider/knowledge_provider.py
@@ -48,7 +48,6 @@ class KnowledgeProvider:
             logger=logger,
             preproc=self.get_preprocessor(kp["details"]["preferred_prefixes"]),
             postproc=self.get_postprocessor(WBMT.entity_prefix_mapping),
-            max_batch_size=1,
             parameters=parameters,
             *args,
             *kwargs,
--- a/strider/knowledge_provider.py
+++ b/strider/knowledge_provider.py
@@ -34,6 +34,7 @@ class KnowledgeProvider:
         kp_id,
         kp,
         logger,
+        parameters,
         information_content_threshold: int = settings.information_content_threshold,
         *args,
         **kwargs,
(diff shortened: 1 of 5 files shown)


strider/knowledge_provider.py
```diff
@@ -47,6 +48,8 @@ def __init__(
             logger=logger,
             preproc=self.get_preprocessor(kp["details"]["preferred_prefixes"]),
             postproc=self.get_postprocessor(WBMT.entity_prefix_mapping),
+            max_batch_size=1,
```
uhbrar, line 51: Have we thoroughly tested how not batching will impact performance?

maximusunc: Good catch. This snuck in cause I have other changes for batch sizing. I've removed it.

uhbrar (approved):

strider/server.py
```diff
@@ -328,17 +330,21 @@ async def sync_query(
     qid = str(uuid.uuid4())[:8]
     try:
         LOGGER.info(f"[{qid}] Starting sync query")
-        query_results = await asyncio.wait_for(
-            lookup(query_dict, qid), timeout=max_process_time
+        # get max timeout
+        timeout = max(
+            max_process_time, query_dict.get("parameters", {}).get("timeout_seconds", 0)
```
kennethmorton, line 335: I forget the complete details for fastAPI query param parsing. Is it possible for timeout_seconds to end up here as a string or some other non-numeric type.

kennethmorton (approved):
Approved, pending comment review in spirit.

strider/server.py
maximusunc: This isn't a query param, it's in a dedicated `parameters` field in the request body on the same level as `message`. And there's no type checking for `timeout_seconds`, so someone could pass whatever they wanted. `parameters` isn't part of the TRAPI spec, so I highly doubt anyone outside of ranking-agent devs are going to even know that this exists.

maximusunc: Added type checks

PR #430, opened 2023-08-23
Query id fix

maximusunc:


strider/trapi.py
```diff
@@ -247,10 +247,58 @@ def filter_message(
...
+                        # need to figure out where query id came from
+                        if node_binding.query_id in curie_map:
+                            # query id is an equivalent identifier that we know about
+                            for eq_id in curie_map[node_binding.query_id].identifiers:
+                                if eq_id in message.query_graph.nodes[qnode_id].ids:
+                                    # found a match
+                                    logger.info(
+                                        f"Changing query_id from {node_binding.query_id} to {eq_id}."
+                                    )
+                                    node_binding.query_id = eq_id
```
uhbrar, line 291: If merging is the reason why this is necessary, maybe instead of changing the query_id to the node_id when they are equivalent, we should get rid of the query_id entirely?

maximusunc: I don't actually check that the node_binding `id` and `query_id` are equivalent, only if either is in the qgraph. Where you put this comment, I'm checking equivalent identifiers of the query_id to see if anything matches in the qgraph. I don't think we can just get rid of the query_id, because we do need it if the kgraph node is a subclass or something.

strider/trapi.py
```diff
@@ -247,10 +247,58 @@ def filter_message(
...
+    """Remove any errant query_ids from result node_bindings."""
+    for result in message.results or []:
+        for qnode_id, node_bindings in result.node_bindings.items():
+            for node_binding in node_bindings:
+                if (
+                    node_binding.query_id is not None
+                    and node_binding.query_id
+                    not in message.query_graph.nodes[qnode_id].ids
+                ):
+                    # if query id exists and isn't in qgraph
```
cbizon, line 272: Does this actually solve the spoke problem?   The message that we send to spoke has DOID:7014 in the query graph.   It sends back a binding with query_id DOID:7014.  So I don't think this block ever gets invoked.

cbizon (changes requested):

strider/trapi.py
maximusunc: This happens in the postproc after we convert the query_graph back to use BMT preferred prefixes, i.e. MONDO, so the query graph id and the `query_id` won't match.

cbizon: Does the knowledge_graph also get converted back at the same time?

maximusunc: It does, along with result node binding ids

cbizon: Would it make sense to also convert query_id?

maximusunc: Good point. I've moved this block up to within `fix_result` and figured out I could simplify the logic a little bit.

PR #432, opened 2023-09-08
Fix duplicate curies issue

maximusunc:


strider/server.py
```diff
@@ -496,6 +501,9 @@ async def lookup(
                 # add new result to hashmap
                 output_results[sub_result_hash] = result
 
+            stop_merging = datetime.datetime.now()
+            message_merging_time += (stop_merging - start_merging).total_seconds()
```
kennethmorton, line 505: I think we should really be using time.time for timing such as these. Not a big deal but FYI

strider/fetcher.py
```diff
@@ -164,6 +164,9 @@ async def generate_from_kp(
             )
         for batch_results in batch(onehop_results, 1_000_000):
             result_map = defaultdict(list)
+            # copy subqgraph between each batch
+            # before we fill it with result curies
+            populated_subqgraph = copy.deepcopy(subqgraph)
```
kennethmorton, line 169: I think another comment or two describing why we need to do this would be helpful. Just describe the problem we are solving right now.

kennethmorton (approved):
Small changes. Take em or leave em.

strider/fetcher.py
```diff
@@ -227,12 +230,17 @@ async def generate_from_kp(
...
                 for qnode_id, bindings in result.node_bindings.items():
-                    if qnode_id not in subqgraph["nodes"]:
+                    if qnode_id not in populated_subqgraph["nodes"]:
                         continue
-                    subqgraph["nodes"][qnode_id]["ids"] = (
-                        subqgraph["nodes"][qnode_id].get("ids") or []
-                    ) + [binding.id for binding in bindings]
-                qnode_ids = set(subqgraph["nodes"].keys()) & set(
+                    # add curies from result into the qgraph
+                    # need to call set() to remove any duplicates
```
uhbrar, line 236: Would this ever cause the curies in the original query graph to be changed? I'm pretty sure that it would stay with the particular subqgraph, but I just want to make sure we wouldn't end up changing the original query graph, since I don't think we want to do that.

uhbrar (approved):

PR #437, opened 2023-11-30
Fix cyclic queries that have subclass results

maximusunc:


uhbrar (approved):
I think this looks good.

strider/fetcher.py
```diff
@@ -272,14 +359,30 @@ async def generate_from_kp(
...
+            self.logger.debug(
+                f"[{qid}] looking for key {key_fcn(subresult)}: {subresult.json()}"
+            )
+            # final hop yield should be empty result
+            if not key_fcn(subresult) in result_map:
+                self.logger.error(
+                    f"[{qid}] Couldn't find subresult in result map: {key_fcn(subresult)}"
+                )
+                self.logger.error(f"[{qid}] subresult from lookup: {subresult.json()}")
+                raise KeyError("Subresult not found in result map")
```
uhbrar, line 383: Instead of erroring out, would it be possible to just log the error and continue, such that if something like this happens, we still get results from the other callstacks?

maximusunc: I'm worried that if we don't do something drastic, we won't end up checking the logs to see this error. We never want this dropping of results to happen.

kennethmorton (approved):
A few questions but mostly suggestions for clarification.

strider/fetcher.py
```diff
@@ -61,14 +62,20 @@ async def lookup(
...
+            qid = str(uuid.uuid4())[:8]
         # if this is a leaf node, we're done
         if qgraph is None:
             qgraph = Graph(self.qgraph)
         if not qgraph["edges"]:
-            self.logger.info(f"Finished call stack: {(', ').join(call_stack)}")
-            yield QueryGraph.parse_obj(
+            self.logger.info(f"[{qid}] Finished call stack: {(', ').join(call_stack)}")
+            # gets sent to generate_from_result for final result merge and then yield to server.py
+            yield KnowledgeGraph.parse_obj(
```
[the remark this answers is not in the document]
maximusunc: Yeah, QueryGraph was always wrong, though it didn't matter because they both have `nodes` and `edges`. If you look down in `generate_from_result`, self.lookup is supposed to yield a subkgraph. I just updated this to return the technically correct type.

PR #439, opened 2023-12-14
Filter and fetcher fixes

maximusunc:


uhbrar (approved):
I think this looks good, however I think this shows that we need to have a more robust test suite. We should consider what cases our current testing fails to cover.

PR #441, opened 2024-01-30
Use blocklist

uhbrar:
Uses the ARS [blocklist](https://github.com/NCATSTranslator/Relay/blob/master/config/blocklist.json) as part of our post processing filtering.


maximusunc (approved):
I tested this with https://github.com/ranking-agent/aragorn/issues/225 and none of the overly generic nodes were in the results.

PR #444, opened 2024-04-15
TRAPI 1.5

uhbrar:
Updates strider to be in compliance with TRAPI 1.5. Also updates URLs to no longer use TRAPI version in URL.

### DIFF
--- a/tests/test_lookup.py
+++ b/tests/test_lookup.py
@@ -83,6 +83,7 @@ async def test_mixed_canonical(monkeypatch, mocker):
                 },
             },
         },
+        False,
         ["infores:kp1"],
         True,
     )
@@ -146,6 +147,7 @@ async def test_symmetric_noncanonical(monkeypatch, mocker):
                 },
             },
         },
+        False,
         ["infores:kp1"],
         True,
     )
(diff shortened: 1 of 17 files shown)


maximusunc (approved):
lgtm

tests/test_lookup.py
```diff
@@ -487,11 +527,77 @@ async def test_node_set(monkeypatch, mocker):
...
+    )
+    QGRAPH = query_graph_from_string(
+        """
+        n0(( ids[] CHEBI:6801 ))
+        n0(( categories[] biolink:ChemicalSubstance ))
+        n1(( categories[] biolink:Disease ))
+        n0-- biolink:treats -->n1
+        """
+    )
+    QGRAPH["nodes"]["n1"]["set_interpretation"] = "ALL"
```
uhbrar, line 595: @maximusunc Can we have this test use a non-set query?

uhbrar (commented):
Besides the test, the changes to the caching looks good.

tests/test_lookup.py
maximusunc: Sure.

PR #447, opened 2024-05-13
Add error handling for cache connection errors

maximusunc:
One thing to note is that if the cache isn't available, Strider will get a new registry of kps for every single query. This is very suboptimal, but would allow us to respond in real time until we could get the redis back up.

### DIFF
--- a/strider/caching.py
+++ b/strider/caching.py
@@ -185,17 +185,13 @@ async def get_kp_registry():
 
 async def get_registry_lock():
     """Lock registry lookup so only one worker will retrieve."""
-    try:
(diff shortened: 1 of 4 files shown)


uhbrar (approved):
I'm assuming that getting the registry isn't really a major issue in terms of performance. If that's the case, this looks fine to me.

maximusunc:
Hmm. I think I can make it better. An in-memory registry that is just loaded on startup. Cause getting it in real time takes about 20 seconds, and that's just cutting into our lookup time.

uhbrar (approved):
I think 20 seconds is a bit more than we want to accept per lookup, so this is much better.

PR #449, opened 2024-06-25
Bump to python 3.12

maximusunc:
A couple smaller things are:
- Switch to a new python profiler. IMO, better than what we had, but still not super consistent with async python.
- Added another condition to post-proc filtering, removes any results that don't match what was asked either by node id or query id.


strider/fetcher.py
```diff
@@ -59,7 +60,7 @@ def __init__(self, logger, bypass_cache, parameters):
         self.bypass_cache = bypass_cache
         self.parameters = {
             **parameters,
-            "batch_size": parameters.get("batch_size") or 1_000_000,
```
uhbrar, line 62: How does this effect kg retrieval? Does it have any effect on response time, response rate, or overall process time?

uhbrar (approved):
I think this looks good overall.

PR #451, opened 2024-07-18
Fix missing kgraph nodes issue

maximusunc:
There are a few issues that led to this:
- the request curie map was getting messed up and returning extra results
- we were adding extra query ids to those extra results
- we now create aux graphs for subclass results
- we weren't adding kgraph nodes for aux graph edges


strider/trapi.py
```diff
@@ -410,6 +414,92 @@ async def fill_categories_predicates(
...
+                                edge_binding.id = primary_edge_id
+
+                    # create subclass edge in knowledge graph
+                    message.knowledge_graph.edges[subclass_edge_id] = Edge.parse_obj(
+                        {
+                            "subject": subclass,
+                            "object": node.id,
+                            "predicate": "biolink:subclass_of",
+                            "attributes": [],
+                            "sources": [
```
uhbrar, line 456: We should be adding aragorn as an aggregator knoweldge source here.

strider/trapi.py
```diff
@@ -410,6 +414,92 @@ async def fill_categories_predicates(
...
+                    if original_edge.subject == subclass:
+                        message.knowledge_graph.edges[primary_edge_id].subject = node.id
+                    else:
+                        message.knowledge_graph.edges[primary_edge_id].object = node.id
+
+                    had_support_graphs = False
+                    for attribute in message.knowledge_graph.edges[
+                        primary_edge_id
+                    ].attributes:
+                        if attribute.attribute_type_id == "biolink:support_graphs":
```
uhbrar, line 478: I don't think we should do this since the attribute_source would be incorrect.

uhbrar (changes requested):

strider/trapi.py
maximusunc: I believe Aragorn is added later on in the Strider process, but I can confirm that Aragorn is listed:
```json
"27b4a949c1a7": {
          "subject": "CHEBI:6149",
          "object": "CHEBI:3215",
          "predicate": "biolink:subclass_of",
          "sources": [
            {
              "resource_id": "infores:automat-hetionet",
              "resource_role": "primary_knowledge_source"
            },
            {
              "resource_id": "infores:aragorn",
              "resource_role": "aggregator_knowledge_source",
              "upstream_resource_ids": ["infores:automat-hetionet"]
            }
          ],
          "attributes": []
        }
```

strider/trapi.py
maximusunc: I'm not sure I understand. This code block is adding the aux graph edge to the the support_graphs attribute of the new edge that has the superclass curie in it. This edge is still being returned by the KP, just in a different format.

uhbrar: Maybe I'm misunderstanding, but I was under the impression that we were checking all of the attributes, and if there already existed an attribute with the "biolink:support_graphs" as the attribute_type_id, then we add the aux graph id to that attribute. If there are no attributes like that, then we create a new attribute, correct?

There should be another "attribute_source" field on that attribute. That field should have "infores:aragorn" listed as the source, but that won't happen in this case if the kp already returned with an attribute that was "biolink:support_graphs".

maximusunc: Added that in.

uhbrar (approved):

PR #452, opened 2024-07-19
Update paths to renci python image

maximusunc:


kennethmorton (approved):
LGTM!

strider/trapi.py
```diff
@@ -492,6 +498,16 @@ def convert_subclasses_to_aux_graphs(
...
+                    message.knowledge_graph.edges[primary_edge_id].attributes.add(
+                        Attribute.parse_obj(
+                            {
+                                "attribute_type_id": "biolink:knowledge_level",
+                                "value": "logical_entailment",
+                                "attribute_source": "infores:aragorn",
+                            }
+                        )
+                    )
+
```
uhbrar, line 510: We need to also remove the existing `knowledge_level` attribute from this edge that would have been created by the KP.

uhbrar (changes requested):

strider/trapi.py
```diff
@@ -452,7 +452,13 @@ def convert_subclasses_to_aux_graphs(
...
                             "object": node.id,
                             "predicate": "biolink:subclass_of",
-                            "attributes": [],
+                            "attributes": [
+                                {
+                                    "attribute_type_id": "biolink:knowledge_level",
+                                    "value": "not_provided",
+                                    "attribute_source": "infores:aragorn",
+                                },
+                            ],
```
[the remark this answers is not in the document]
maximusunc: Added.

strider/trapi.py
maximusunc: Now we'll overwrite knowledge_level if it's present, otherwise add a new attribute

uhbrar (approved):
This looks good now. KL/AT are both required attributes, so if they are not present for an edge when receiving it from a KP, we should probably log it and set it to `not_provided` so that we can then post an issue for that KP. That's beyond the scope of this PR though.

PR #456, opened 2024-08-05
Change kp refresh function name and pin redis version

maximusunc:
The watchdog tests have been failing for a little while because the /kps endpoint was returning a 500. This fixes that.

### DIFF
--- a/strider/server.py
+++ b/strider/server.py
@@ -74,7 +74,7 @@ openapi_args = dict(
     title="Strider",
     description=DESCRIPTION,
     docs_url=None,
-    version="4.7.2",
+    version="4.7.3",
     terms_of_service=(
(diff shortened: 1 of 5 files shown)


strider/server.py
```diff
@@ -186,7 +186,7 @@ async def catch_exceptions_middleware(request: Request, call_next):
 
 
 @APP.on_event("startup")
-async def get_kp_registry():
+async def refresh_kp_registry():
```
uhbrar, line 189: If we change the name of this function, what are we calling now in line 682?

maximusunc: We are now correctly calling line 48

uhbrar: Ah okay, I missed that import. Thanks.

uhbrar (approved):
lgtm

PR #459, opened 2024-08-21
Mcq urls

maximusunc:


strider/config.py
```diff
@@ -27,6 +27,14 @@ class Settings(BaseSettings):
     use_cache: bool = True
     offline_mode: bool = False
 
+    # MCQ KP URLS
+    mcq_answer_coalescer: AnyUrl = "https://answercoalesce.renci.org/query"
+    mcq_genetics: AnyUrl = (
+        "https://translator.broadinstitute.org/genetics_provider/trapi/v1.5/query"
+    )
+    mcq_cohd: AnyUrl = "https://cohd.io/api/query"
+    mcq_semsemian: AnyUrl = "http://mcq-trapi.monarchinitiative.org/1.5/query"
```
uhbrar, line 36: Can we do this via the smartapi metadata instead? It looks like the KPs have `multicuriequery: true` in the `x-trapi` block of their smartapi listings.

maximusunc: This is news to me

maximusunc: No, it is not standard.

maximusunc: Semsemian isn't even registered on SmartAPI

uhbrar: The decision was made on architecture that any service that provides MCQ should have that labelled. It looks like genetics and COHD at least do, while the other two don't. We should encourage them to do as such, since hard coding URLs isn't the greatest solution. But, if it's all we have for now, then there's nothing we can do about it.

uhbrar (approved):

